PR #10, opened 2023-04-16
add tuxguitar-tray in swt configuration

guiv42:
Just a suggestion: proposal to integrate tuxguitar-tray module in swt configurations. In fact I was used to it in my 1.5.4 config, but I just discovered it had been added by a patch (see [here, line 148](https://build.opensuse.org/package/view_file/openSUSE:Factory/tuxguitar/tuxguitar.spec?expand=1)) during the rpm packaging phase.
Notes:
- it is a bit surprising: this tuxguitar-tray module does not seem to be used anywhere
- there is a swt dependency, so it is not relevant in qt or jfx configurations
- I could not test it in freebsd, debian, macos and windows, but logically it would make sense to apply the same modification to corresponding poms

There is another module, tuxguitar-viewer, also introduced in TG1.5.4 rpm by a patch during packaging. I don't know what it corresponds to, so I have not addressed this question.


guiv42:
Hum.. forgot ".jar" extension in pom file, I need to correct that

PR #17, opened 2023-04-22
translate plugins-related menu items in "Tools" menu

guiv42:
Hi @helge17. This is my very first merge conflict, it seems we both tried to address this translation issue in parallel!
I think your implementation does not enable the labels to be updated when the language setting is changed (minor issue). Could you please try the following sequence?
- look at labels in "Tools" menu
- change language
- look again at labels: I think they should not be updated
- after closing / re-opening TG it should be OK

Another point: 3 plugins can be displayed in the Tools menu. It seems your implementation does not translate "Jack Console" if the corresponding plugin is activated in the settings.

Anyway, it's up to you to decide which direction to follow between these 2 options!


helge17:
Thank you! You are definitely the better coder than me ;-)

guiv42:
You're welcome :)
Take care, I'm no software engineer, just a hobbyist and very capable of doing bad things! I hope others join, this could help code reviews, it's always safer when several people have a look at such an evolution.
Regarding this specific issue, the best solution might be somewhere in-between, as often.
Definitely it was needed to store plugins-related menuItems in ToolMenuItem class, or it would have been completely impossible to refresh the labels when language changes.
On the other hand I am not 100% convinced by the "label" property I have added in TGCustomTool - and that you haven't. Might be an overkill... I preferred that to avoid interfering with possible other uses of this class, which I have not analyzed.
I will have a detailed look, and possibly send another pull request for simplification if it's safe to delete it.

helge17:
I can't really analyze the pull requests in depth, but I test to see if they do what they are supposed to do, and your fix worked fine for me. I'm also not a software developer.

Looking at all the patches on SF.net and the 2.0beta fork, I think there are some people out there with programming skills who want to improve TG. So let's hope they join our new initiative.

PR #18, opened 2023-04-23
deletion / exclusion of eclipse related files

guiv42:
Playing with a text editor and a console is only fun for the first few hours, so I finally imported the project in Eclipse.

This created a lot of hidden files (starting with ".") all over the different sub-folders of the project.
None of these files makes sense in the git repo, so I suggest updating the .gitignore file accordingly.

This also showed that (probably for historical reasons) some of such files were already present in TuxGuitar-debug-helper and TuxGuitar-editor-utils. I suggest deleting them.


helge17:
I guess also these files could be deleted(?)
- TuxGuitar-android-midimaster/.project
- TuxGuitar/.project
- TuxGuitar-debug-helper/.project
- build-scripts/.project
- TuxGuitar-editor-utils/.project

guiv42:
Yes, I think you can delete them. In my Eclipse project I deleted them all without any problem.  This test does not cover all cases, as I only build the linux-swt-x86_64 in the Eclipse project. But I got a warning about the Tuxguitar/.project file... and  it automatically recreated the same file with the same content. So this should be safe.

Issue #20, opened 2023-04-29
cannot build Qt5 version


guiv42:
This post summarizes my attempt to build the Qt5 version of TuxGuitar (that I would like to patch).
There is a TODO file in build-scripts/tuxguitar-linux-qt5-x86_64/, stating that qtjambi should be [downloaded](https://github.com/tilialabs/qtjambi5) and built from source. Instructions given on this site about the procedure to build seem a bit confusing for me.
Furthermore, there are several versions of qtjambi out there. The one referred to by the TODO file seems a bit outdated (2017), and the latest release seems to be [there](https://github.com/OmixVisualization/qtjambi).
So I downloaded:
- the last available version of qtjambi for Qt5 from [here](https://github.com/OmixVisualization/qtjambi/releases/tag/v5.15)
- the last available version of qtjambi-native-linux-x64 for Qt5 from [here](https://repo1.maven.org/maven2/io/qtjambi/qtjambi-native-linux-x64/5.15.14/qtjambi-native-linux-x64-5.15.14.jar)

I installed both in maven environment following instructions from TODO file:
```
mvn install:install-file -Dfile=qtjambi-5.15.jar -DartifactId=qtjambi -Dpackaging=jar -DgroupId=org.qtjambi.qt -Dversion=5.15
mvn install:install-file -Dfile=qtjambi-native-linux-x64-5.15.14.jar -DgroupId=org.qtjambi.qt -DartifactId=qtjambi-native-linux64-gcc -Dversion=5.15 -Dpackaging=jar
```

I upgraded qtjambi version from 5.5.1 to 5.15 in the following pom files:
- build-scripts/tuxguitar-linux-qt5-x86_64/pom.xml
- TuxGuitar-ui-toolkit-qt5/pom.xml

And when I try to build tuxguitar-linux-qt5-x86_64, I get a lot of errors:
- "package xxx does not exist"
- "cannot find symbol: class xxx"

So it seems I have done something wrong when trying to set up qtjambi

note: I have not yet addressed the question to move to Qt6

guiv42:
Update
[...]

PR #21, opened 2023-04-30
attempt to solve shortcut issue with non-EN keyboard

guiv42:
Description of the issue
------
Some users have problems using keyboard shortcuts to insert new notes in the tab. The problem is present with keyboards:
- with no numpad (typ. compact laptop)
- at least AZERTY layout, possibly others

See https://sourceforge.net/p/tuxguitar/discussion/522985/thread/84fa6888dc

On these keyboards, digits are only reachable with a combination of keys, e.g. "shift + &" = "1".
It's still possible to redefine the keyboard shortcuts through dedicated setting dialog manually, but this is not very intuitive. When displayed shortcut is "1", user expects to trigger this shortcut when typing "1", whatever the keys combination used. Especially when only one such combination exists!
Furthermore, non-en keys are not correctly displayed in this dialog. E.g. "shift + à" corresponds to "0", but is displayed "shift + undefined" in JFX variant. So, this shortcut settings dialog is not considered an acceptable solution.

Workaround
----
I tried a fix with a generic behavior for all types of keys, but it's impossible to anticipate all combinations for all keyboards, some of them even leading to non-printable characters.

So the proposed modification is more a *workaround* than a *fix*, anyway it is minor and very localized, so it should be harmless. Basically, when one keyboardReleased event is detected, if the entered text is a digit, then this digit overrides the combination of keys used to enter it, and "shift" modifier is ignored.
Only SWT and JFX variants are handled, not yet Qt (because of issue #20). Anyway the same behavior should be easily implementable in Qt variant, where the additional required event attribute is also available.

Test
-----
I only tested it in my AZERTY configuration, and only in Linux SWT and JFX configurations.
A short regression test on a more "conventional" keyboard layout (qwerty, or any layout where first line enables direct access to digits without "shift" modifier) would be useful I think before eventually merging this pull request

regards,
guiv

### DIFF
--- a/TuxGuitar-ui-toolkit-jfx/src/org/herac/tuxguitar/ui/jfx/resource/JFXKey.java
+++ b/TuxGuitar-ui-toolkit-jfx/src/org/herac/tuxguitar/ui/jfx/resource/JFXKey.java
@@ -87,11 +87,25 @@ public class JFXKey {
 	}
 	
 	public static UIKeyCombination getCombination(KeyEvent keyEvent) {
+		/* workaround for non-EN keyboards without numpad (typ. azerty laptop)
+		   with these keyboards, digits can be entered only with key combination, e.g. "shift+&" = "1"
+		   Shortcuts configuration dialog shows "1" as an active shortcut, so when user types "shift+&" this shall be considered "1"
+		   but, it's not possible to consider the typed character in all cases (even a non-printable character)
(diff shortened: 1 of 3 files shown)


helge17:
I tested the patch on combinations of SWT & JFX, Windows & Linux, german QWERTZ keyboards, laptops with and without numeric keypad. I did not notice any change, the numbers work on both the keyboard row and the numeric keypad, with and without the patch.

Issue #31, opened 2023-05-12
impossible to add new parameters in tuxguitar.cfg file in the future


guiv42:
This is a side effect of commit f5fafce1e8b591f898e3c73f1b983857d555e7c9

Since the folder name for user configuration doesn't contain TuxGuitar's version any more, new versions SHALL be able to load config files created by old versions. Or it will never be possible to add new config parameters in the future.
Currently this situation is not correctly handled.
While working on a new feature, I tried to add a new parameter. When starting TuxGuitar with an old version config file on disk (i.e. without any value for this new parameter), I expected the new parameter to take its defined default value. But yet the new parameter is not set (remains `null`).

See [TGConfigPropertiesHandler.java](https://github.com/helge17/tuxguitar/blob/f5fafce1e8b591f898e3c73f1b983857d555e7c9/TuxGuitar/src/org/herac/tuxguitar/app/system/config/TGConfigPropertiesHandler.java). There is an attempt to read first default values, then to override them with values read from stored config file:
```
public void readProperties(TGProperties properties, String module) throws TGPropertiesException {
	this.readDefaultProperties(properties, module);
	this.readStoredProperties(properties, module);
}
```
Unfortunately this does not work, the first call to `readDefaultProperties` is ineffective: when  `readStoredProperties` is called it ends up in [TGFilePropertiesReader.java](https://github.com/helge17/tuxguitar/blob/f5fafce1e8b591f898e3c73f1b983857d555e7c9/TuxGuitar/src/org/herac/tuxguitar/app/system/properties/TGFilePropertiesReader.java), where the previously loaded defaults are cleared!
```
if( file.exists() ){
	((TGPropertiesImpl)properties).clear();
	((TGPropertiesImpl)properties).load(new FileInputStream(file));
```

So, the stored properties do not *override* default values, but *overwrite* them with the content of old config file, leaving new parameter to `null`. At one point this will necessarily propagate to a null pointer exception...

Edit: a simple way to reproduce without instrumenting code
delete "color.play.note=255,0,0" line from tuxguitar.cfg (stored in ~/.config/tuxguitar/config on Linux), start TuxGuitar and open menu Tools/Settings. This raises an exception

helge17:
The pull request #32 needs to be extended to the Android version. Currently the APK does not build.

helge17:
Thank you!

PR #32, opened 2023-05-13
Fix for issue #31

guiv42:
when reading a conf file from an older TuxGuitar's version, not all parameters may be defined. Undefined parameters need to be set to their default values.

Before this fix, undefined parameters led to null TGProperties objects 

see #31


helge17:
With the version number in the name of the config folder, all user settings were gone every time TuxGuitar was updated. By deleting the version number I wanted the users to keep their settings after an update.

But I have to admit that I didn't test what happens when new settings are added.

Your patch takes into account new entries in tuxguitar.cfg, but I'm not sure if it also reads new entries from tuxguitar-plugin-settings.cfg. Shouldn't this file be taken into account as well?

Sorry for the extra work... I can also put the version number back into the config folder name, then TuxGuitar would at least behave like version 1.5.6.

guiv42:
I'm perfectly OK with this removal of version number in folder name. I've read a bit on SF the history of this patch, it's clear that successive versions somehow polluted the users' home folder. So no problem about that, we just have to do it properly.
Thank you for completing the analysis: I have not tried for plugin settings config file, essentially because I don't know what it's used for. No idea if my patch would handle this case also...
If you know one feature which is using some of these parameters it should be quite easy to test: just edit the config file, remove manually a few lines, start TuxGuitar and play with this feature (see the example in #31 with color.play.note and settings dialog). It may or may not raise an exception, depending from each feature implementation...
Tell me if you can identify such a scenario, I could test and debug if needed.

helge17:
I made the following tests with TuxGuitar including this patch:

1. rm -rf ~/.config/tuxguitar && tuxguitar &
2. Look at Tools - Plugins and verify:
   Debug Helper Plugin:              disabled
   TuxGuitar Community Integration:  disabled
   Java Sound Api Plugin:            enabled
3. Close Dialog and exit TuxGuitar without changing any setting.
  ~/.config/tuxguitar/config/tuxguitar-plugin-settings.cfg contains the default settings:
  tuxguitar-community.enabled=false       # OK, as observed above
  tuxguitar-jsa.enabled=true              # OK, as observed above
  tuxguitar-fluidsynth.enabled=false
  tuxguitar-debug-helper.enabled=false    # OK, as observed above
  tuxguitar-jack.enabled=false
  tuxguitar-synth-export.enabled=true
  tuxguitar-gervill.enabled=false
  tuxguitar-alsa.enabled=false
4. Remove the three lines from the file. The other setting remain:
   tuxguitar-fluidsynth.enabled=false
   tuxguitar-jack.enabled=false
   tuxguitar-synth-export.enabled=true
   tuxguitar-gervill.enabled=false
   tuxguitar-alsa.enabled=false
5. tuxguitar &
6. Look at Tools - Plugins and verify:
   Debug Helper Plugin:              enabled    # not the default setting
   TuxGuitar Community Integration:  enabled    # not the default setting
   Java Sound Api Plugin:            enabled    # default setting, OK
7. Close Dialog and exit TuxGuitar without changing any setting.
8. ~/.config/tuxguitar/config/tuxguitar-plugin-settings.cfg still the same as in 4.
9. cat /dev/null > ~/.config/tuxguitar/config/tuxguitar-plugin-settings.cfg
10. Tools - Plugins: All plugins enabled

=> If a new plugin is added it is always enabled, even if it should be disabled according to /opt/tuxguitar/dist/tuxguitar-plugin-settings.cfg .

But this case is very rare, and TuxGuitar does not crash. So I think it is better to keep the version number out of the config folder name and add your patch.

guiv42:
Thanks for this detailed tests!

Issue #28, opened 2023-05-04
Tracklist/song played measure position scrolling during playback is worse than 1.2


kainah:
I'm sorry if that change of TuxGuitar's behaviour has been discussed earlier, maybe on Sourceforge.
 
When TuxGuitar 1.2's playing position (currently playing measure) in the tracklist gets to the edge of the tracklist area (forwards or backwards), it moves the tracklist horizontally by it's whole width.
However, in the latest snapshot (2023-05-02), and maybe since 1.3 the tracklist scrolls by a single measure forwards or backwards. This is quite uncomfortable if you try to quickly move around the track while playing.
Is there anyone else who also thinks that's a problem?

To be precise, I'm attaching screen captures of 1.2 tracklist behaviour and latest snapshot's one. SWT and JavaFX versions of the new TuxGuitar are identical in that regard.

https://user-images.githubusercontent.com/1732918/236137906-d5aa8a67-f9c4-4f53-8f66-3cb5856fd1e9.mp4

https://user-images.githubusercontent.com/1732918/236137929-30847a7e-1f99-48e2-8613-1916259dee52.mp4

guiv42:
Hi
I agree that scrolling one measure at a time is not very comfortable, because you lose the ability to anticipate.
On the other hand, scrolling the tracks by the whole width is a kind of discontinuous behavior, where you lose instantly all what you could see before it scrolled.
Don't know if what I say is very clear... And thanks @kainah for the short videos, this definitely helps understanding!

What about a possible intermediate behavior, where scrolling could be done by 3/4 of the tracks width? Or [full width minus a few measures]?
In this case, when horizontal scrolling is triggered, the currently playing measure would appear on the left part, but not fully left aligned: the few preceding measures would still be visible.
It could also be possible to trigger the horizontal scrolling a few measures before the edge is reached (still for anticipation)

Note: this proposal only makes sense when scrolling *forward* while playing.
@kainah I do not understand the case you mention moving *backwards*?

kainah:
@guiv42 Yeah, that kind of horizontal scrolling with a few measures "left behind" after scroll would be even better than 1.2! Didn't think of it myself.
[...]

PR #35, opened 2023-05-17
smooth horizontal scrolling of tracks table when playing

guiv42:
This is a proposal for an evolution of tracks table horizontal scrolling, with the objective to address issue #28 

The idea is:
- when playing, to anticipate horizontal scroll of tracks table by a few measures for a better anticipation
- when scrolling, leave a few measures "behind" for continuity

On a wide screen it will leave 3 measures margin left and right. This margin is reduced if the window is so small there is not enough space for that.
This margin is only applied when playing: when *not* playing, moving with arrows or mouse should have the same behavior as before (scroll measures 1 by 1)

### DIFF
--- a/TuxGuitar/src/org/herac/tuxguitar/app/view/component/table/TGTableViewer.java
+++ b/TuxGuitar/src/org/herac/tuxguitar/app/view/component/table/TGTableViewer.java
@@ -442,7 +442,7 @@ public class TGTableViewer implements TGEventListener {
 				this.resetTexts = false;
 			}
 			if( this.followScroll ){
-				this.followHorizontalScroll(getEditor().getTablature().getCaret().getMeasure().getNumber());
+				this.followHorizontalScroll(getEditor().getTablature().getCaret().getMeasure().getNumber(), 0);
 				this.followScroll = false;
 			}
 			getControl().redraw();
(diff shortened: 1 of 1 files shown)


ckolumbus:
Hi, is this functionality also available in the android version? I could not find a way how to enable  horizontal scrolling in the android app.

guiv42:
Hi
no this functionality only concerns the "tracks table" (see the bottom part of screen in videos included in #28), which is not displayed by the Android version

ckolumbus:
Could you give me maybe some hint where to add this to the "page" view scrolling, I'd REALLY love to have this on my tablet for gigs.

guiv42:
If you want to dig in the Android version it will be welcome ;)
There's a lot of work to do on this app (upgrade dependencies, many warnings to fix, etc.)

To be honest, since this fork started a few months ago we essentially focused on the desktop version (just a matter of available time). But at least we took care to maintain full compatibility, so Android app still builds and works as before.

So unfortunately I don't think I can really help you.

ckolumbus:
I'm not familiar with mobile dev either, but I'd like to have this feature anyway in both apps. I hope that I'll find some time to dig into this .. thanks for the pointers to the code

PR #43, opened 2023-05-22
proposal: add a control in toolbar to show/hide tracks table

guiv42:
this is useful when working with a small screen
feature essentially back-ported from 2.0b fork

notes:
- new strings to be translated
- icons are also imported from 2.0b fork

tested on Linux (OpenSuse tumbleweed) and Windows 10, in both SWT and JFX configurations
Android: I only checked that it still builds and runs, it should have no impact


helge17:
Looks useful, thanks! Works great using the icon in the main toolbar. But I miss an entry for this feature in the view menu.

guiv42:
Yes that makes sense, I will try to add that

PR #49, opened 2023-06-03
solo/mute icons in trackTable

guiv42:
starting to backport (slowly) trackTable improvements from 2.0beta branch.
- added solo/mute icons in trackTable, instead of simple label "S", "M",...
- deleted corresponding strings in translation files (were not translated anyway)
- copied icons from 2.0beta (png only, not svg)

note: also ported to JFX configuration
tested on Linux (openSuse tumbleweed), Window (win10) in both SWT and JFX configurations. Android version unchanged


helge17:
While playing around with the 2.0beta theme, I noticed that when you change the theme, the solo/mute icons are only updated after restarting TG.

guiv42:
Thanks for noticing that point, I've git to fix that.
Would you mind creating a dedicated issue so I don't forget ? I don't know when I will be able to do that but for sure not immediately

PR #50, opened 2023-06-03
cleanup

guiv42:
just deleted unused imports (were present in 2.0beta)
should  have seen that before #49...


guiv42:
Also added markers in the trackTable
but using icons instead of the triangles from 2.0b
markers_Oxygen
markers_Breeze

name of the pull request would need to be adapted... it's not only "cleanup"
Is there a way to do this?

PR #59, opened 2023-06-26
feature: use custom template for new song creation

guiv42:
Attempt to implement feature request: https://sourceforge.net/p/tuxguitar/feature-requests/102/
When a user template is selected, all "new song" actions create an empty song from the defined template: menu file/new, toolbar, "Ctrl+n" shortcut, closing last opened file, or simply starting TG.

Feature is accessible through menu "File/Custom template"

Initiating merge request, but feature is not 100% mature:
- a few complementary tests would be appreciated
- also a feedback on the user interface
- in "Custom template" dialog, a recycle bin icon would probably be more suitable than the 'minus' sign I re-used from existing icons. But I do not know how to create such an icon (for all skins)

See commit message for detailed description of modification and of performed tests.

New dialog for custom template selection:
custom-template-dialog


helge17:
The feature looks useful, thanks! I would suggest moving it to the menu, like:
```
  File → New:
     New Song
    +---
    +Set custom Template
    +Revert to default Template
     ---
     Blues Trio
     Classical Guitar Duo
```
This way it's easier to understand what it does, and you don't have to worry about the menu icons (the skins need to be reworked anyway).

guiv42:
Thank you for your feedback, I like the idea to simplify and delete the dialog. Dialog may come back one day if we provide a more elaborated feature, enabling user to define several templates (it was my initial intention). Okay, first step, keep it simple.
I'm a bit more concerned about the location of the 2 menu items "select/delete template". In a level-2 sub-menu I fear it might become unnoticed, creating de facto a "hidden" feature. I personally never open "File/New": Ctrl+N and toolbar icon are more easily accessible.
So, what is your opinion about following propositions?
1. delete new Dialog
2. replace current menu item "File/Custom template" by 2 items "select"/"delete" at the same level, with horizontal separators above and below
3. prepare future enhancement: move imported user template from [user conf folder]/config to [user conf folder]/config/templates. This mimics TG's config internal structure
4. for robustness: if new song creation from user-defined template fails (corrupted file), delete it. Back to a clean situation (with a corrupted template, delay to fallback to default is really long on my old Win machine)

guiv42:
OK, that sounds good
Update coming soon...

guiv42:
This should be OK.
also checked that it works if no user folder config exists (1st install).
tested on Linux (OpenSuse tumbleweed) & Win10, with both variants JFX and SWT.
Android: only tested it still builds, logically this should have no impact

Issue #62, opened 2023-07-06
Android error on „Delete Note or Rest“: „Attempt to invoke virtual method ‚boolean org.herac.tuxguitar.util.TGNoteRange.isEmpty()‘ on a null object reference“


helge17:
When tapping on the „<“ sign (Delete Note or Rest) near the „6“ the error message „Attempt to invoke virtual method ‚boolean org.herac.tuxguitar.util.TGNoteRange.isEmpty()‘ on a null object reference“ is shown. Error is 100% reproducable.
„>“ (Insert Rest Beat) beside the „9“ works as expected.
Last working commit: Thu May 18 19:55:09 2023 +0200 fa63820c8d69a5277535abdc73eb1b9c5bd0df6d
Introduced with commit: Fri May 19 00:30:27 2023 +0200 e12be785182856a96059bfd90579d5678d6c0909

![grafik](https://github.com/helge17/tuxguitar/assets/5897631/cfe47190-f188-48bf-9e08-debfee290b96)

guiv42:
Yes, sorry for that.
Not a big surprise in fact: I just detected the very same issue in other selection-related actions I am currently trying to backport (e.g. TGSetDurationAction). It seems 2.0beta fork does not maintain full compatibility with Android version. I just checked, logically this should be the only action yet backported with this issue.
It should be easy to fix (pull request coming soon).
It just means Android compatibility needs to be more carefully checked when backporting any other feature from 2.0b...

helge17:
Works great now, thanks!

PR #65, opened 2023-07-07
apply actions to selection: set note duration, clean beat

guiv42:
essentially backported from 2.0beta fork, but adapted to keep compatibility with Android version:
if the NOTE_RANGE and/or BEAT_RANGE attributes are not defined in action context or are empty,
then keep original implementation
Tested on Linux & Win10, JFX and SWT. This time I also tested the absence of regression in Android config (see #62)


guiv42:
Warning: not ready for a merge. Just detected it does not work correctly if selection is spread over several measures.
Correction in progress

guiv42:
that should be better

PR #75, opened 2023-07-26
Apply "shift up/down" action to selection

guiv42:
Action accessible through menu item "Beat / Shift up(down)" 
Detailed implementation is a bit tricky, I would not be surprised if some bugs remain present (hopefully minor).
Only tested on Linux platform

Improvements compared to 2.0beta fork implementation:
- fix: shift only by 1 string. For fun, just try to "shift down a string" a D major chord x.x.0.2.3.2 with 2.0b: you get 22.21.0.2.x.x !
- fix: undo/redo also works if selection covers several measures
- fix: menuItem enabled if something is selected, even if cursor is not on a note
- when applied to selection, shift either all selected notes or none of them
- maintained compatibility with Android (unchanged behaviour)

behaviour if nothing is selected remains unchanged: shift 1 single note under cursor
note: it shifts note not necessarily by 1 string, but *to the next free string*
I'm not convinced this is the best solution


helge17:
Thanks for the patch! I noticed that the beat where the cursor stops is always played, even if the selection goes over multiple beats. Shouldn't the beat only be played when only one beat is selected (as in #67)?

guiv42:
Yes, clearly you're right, thanks for the feedback. As mentioned above I did not expect this evolution to be perfect!
In fact it just means I have not completely understood how the context updated by action is used, and sometimes plays selected beat. Most probably I will need to rework both features (this pull request and #67).
Update coming soon...

PR #82, opened 2023-07-28
initialize start/end of training mode with selected measures

guiv42:
to test:
- select some measures
- type F9 (or menu Player/play mode)
loop range is initialized with measures from selection

### DIFF
--- a/TuxGuitar/src/org/herac/tuxguitar/app/view/dialog/transport/TGTransportModeDialog.java
+++ b/TuxGuitar/src/org/herac/tuxguitar/app/view/dialog/transport/TGTransportModeDialog.java
@@ -6,6 +6,7 @@ import java.util.List;
 import org.herac.tuxguitar.app.TuxGuitar;
 import org.herac.tuxguitar.app.action.impl.transport.TGTransportModeAction;
 import org.herac.tuxguitar.app.ui.TGApplication;
+import org.herac.tuxguitar.app.view.component.tab.TablatureEditor;
 import org.herac.tuxguitar.app.view.controller.TGViewContext;
 import org.herac.tuxguitar.app.view.util.TGDialogUtil;
 import org.herac.tuxguitar.document.TGDocumentContextAttributes;
@@ -29,6 +30,7 @@ import org.herac.tuxguitar.ui.widget.UIRadioButton;
 import org.herac.tuxguitar.ui.widget.UISelectItem;
 import org.herac.tuxguitar.ui.widget.UISpinner;
 import org.herac.tuxguitar.ui.widget.UIWindow;
+import org.herac.tuxguitar.util.TGBeatRange;
 
 public class TGTransportModeDialog {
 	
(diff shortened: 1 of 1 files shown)


helge17:
Thanks for the patch! I think the cursor should be set to the first measure in the range as soon as the play button is pressed; currently the measure where the cursor stops (usually the last in the selection) is played once before the loop starts. And maybe it makes sense to also set the "Play looped" switch automatically?

guiv42:
Thank you, I really appreciate having some feedback.
There are two different questions here:
1. cursor position
You're right it does not make sense to start playing with the last measure. However, moving cursor to beginning of range *when "play" button is pressed* does not seem a good solution to me. Basically, "play" button starts playing where the cursor is, and I think this should stay like this, or it would break the ability to pause somewhere - then restart from there (see 421a0a3773918c2a29cd7465b4f51fe0b2a87689).
Another possible option could be to move cursor to the first beat of range for training mode *when the "OK" button in player mode dialog is clicked*? (did not check yet if it was feasible)
2. "play looped" automatically
I'm not sure I have fully understood your proposal. Does that mean to check by default the "play loop" checkbox in "simple mode"? If so, I have no idea about what's best, it really depends how people use the software. Do you know how other music players behave by default?

helge17:
1. I'm not exactly sure what is the correct technical solution, but i compared the two ways to set up a loop:

A. Loop without selection
  - Place cursor in measure 2
  - F9 (Play Mode)
  - Select "Play looped"
  - Set "Loop Range" from measure 8 to 10
  - OK (cursor remains in measure 2)
  - Click Play
  - Measure 2 is not played, only 8->9->10->8... as expected
  - Click Stop: Cursor always remains in measure 8 (even if you stop in measure 9 or 10)
  - Click Play again: Playing always starts at mesure 8

(close and reopen TuxGuitar)

B. Loop with selection
  - Select measures 8-9-10 (cursor stops in measure 10)
  - F9 (Play Mode)
  - Select "Play looped"
  - "Loop Range" is automatically and correctly set to measure 8 to 10
  - OK (cursor remains in measure 10)
  - Click Play
  - Playing starts at measure **10**->8->9->10->8...
  - Click Stop: Cursor always remains in measure **10** (even if you stop in measure 8 or 9)
  - Click Play again: Playing always starts at measure **10**

A. works as expected, B. doesn't.

2. My old Guitar Pro 6 behaves as follows:

  - If you press Play without selection, playback starts at the point where the cursor is and ends at the end of the song.
  - If you select some measures and then press Play, the selected measures are played once.
  - If loop mode is enabled and some measures are selected, the selection will play in a loop when you press Play.

Maybe it would be a good idea to always play the selected measures, even if you select Play immediately without using F9 first. But TuxGuitar supports playing part of the song only in loop mode, not just once, so it might be feasible to always select "Play looped" when there is a selection. To implement GP6's behavior, the "Play Mode" dialog would need some redesign.

But I have to admit that TuxGuitar has some more important problems than this ;-)

guiv42:
It would be easy to set up automatically "play looped" when typing F9 with a selected area, but I agree it would make sense to be able to play selected area directly, even without typing F9.

The problem is that there is currently no link in the software between the selection and the player mode. You can reproduce exactly the scenario B without selecting anything: just place cursor in measure 10, type F9, configure a loop with measures 8 to 10, then Play: it plays 10 -> 8 -> 9 ...

What I understand from current implementation is that cursor is used to store the place where Play should (re)start. At least if the cursor is inside the loop range (I've got to analyze scenario A). Player does not store any variable to make a difference between paused and stopped states.
Looking at scenario B: select measures 8-9-10, (cursor stops in measure 10)
At that point, the player does not know *why* the cursor is in measure 10
- if it's there because user just selected the range, it should start play from measure 8
- if it's there because user paused player there, it should (re)start play from measure 10

Got to think of this interaction between selection and player globally... And I agree with you to consider this low priority.

guiv42:
After analysis, I was wrong about 1 thing: Player does store an internal reference position (tickPosition), independent from cursor position. These 2 positions are aligned following some events. Non-exhaustive list:
- when player pauses: cursor is moved (421a0a3773918c2a29cd7465b4f51fe0b2a87689)
Note: cursor is *not* moved when player starts or stops.
- when user moves cursor: tickPosition is moved. If a loop is defined then tickPosition is moved to the cursor's closest measure within loop interval

That explains both scenarios A and B. If a loop 8-10 is defined and cursor is moved to measure 12, then play will also start with measure **10**, just as scenario B.

When user moves cursor *while selecting with click&drag* that should be fairly easy to move tickPosition in consistency with *selection start* instead of *cursor position*. This should at least solve scenario B. Just a few tests to do, then I may submit a new pull request.

I'm also considering a rework of player modes to take selection into account in a cleaner way. If I can get something working I will address this in a dedicated discussion.

Issue #88, opened 2023-08-10
Android: erroneous title of Harmonic dialog


guiv42:
Select one note, then click on menu item "Effects/Harmonic"
Options correctly correspond to harmonic effect, but title is "Grace note"

PR #90, opened 2023-08-10
correction of harmonic dialog (Android)

guiv42:
fix for #88 
in many languages: copied text from action_change_harmonic to harmonic_dlg_title
also corrected a few errors in french version

### DIFF
--- a/TuxGuitar-android/res/values-b+bg/strings.xml
+++ b/TuxGuitar-android/res/values-b+bg/strings.xml
@@ -435,7 +435,7 @@
 	<string name="grace_dlg_transition_hammer">Форшлаг</string>
 
 	<!-- Harmonic dialog -->
-	<string name="harmonic_dlg_title">Фиоритура</string>
+	<string name="harmonic_dlg_title">Флажолет</string>
 	<string name="harmonic_dlg_type_nh">[N.H] Естествен флажолет</string>
 	<string name="harmonic_dlg_type_ah">[A.H] Изкуствен флажолет</string>
 	<string name="harmonic_dlg_type_th">[T.H] Флажолет с потупване</string>
--- a/TuxGuitar-android/res/values-b+ca/strings.xml
+++ b/TuxGuitar-android/res/values-b+ca/strings.xml
@@ -435,7 +435,7 @@
 	<!-- <string name="grace_dlg_transition_hammer"></string> -->
 
 	<!-- Harmonic dialog -->
-	<string name="harmonic_dlg_title">Appoggiatura</string>
+	<string name="harmonic_dlg_title">Harmònic</string>
 	<string name="harmonic_dlg_type_nh">[N.H] Harmònic</string>
 	<string name="harmonic_dlg_type_ah">[A.H] Harmònic artifical</string>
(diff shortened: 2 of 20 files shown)


helge17:
All Android translations are derived from the desktop translations using the script misc/messages_desktop_to_android.pl (see commit ed75885038f45c5b4b42eeca79bdf5e0e1e6504d). -h shows a brief explanation of how it works. The script is far from perfect, but it takes too much time to translate the desktop and Android versions separately. One of the problems with adopting the desktop translations is that the variable names in Android differ from the desktop version (e.g. action_track_add in Android is called action.track.add-new in the desktop version). So in the long run it makes sense to clean up and align the translations on both platforms, but for now I think the results of the script are acceptable.

guiv42:
Thank you!
I did not try the script yet. Is there a way to list all Android string names which are not filled by this script? It might be possible to refactor the Android resources files to make names match the desktop version...

helge17:
The script writes the file /tmp/keys_android-keys_desktop-values_android.csv where you can see how the variable names are matched:

```
# keys_android                                               keys_desktop                                                 values_android
app_name                                                     UNKNOWN                                                      TuxGuitar
action_settings                                              action.gui.open-settings-editor                              Settings
action_settings                                              settings                                                     Settings
action_settings                                              tools.settings                                               Settings
global_error_message                                         UNKNOWN                                                      An unexpected error has occurred
global_exit_confirm_message                                  UNKNOWN                                                      Are you sure you want to exit?
global_button_ok                                             ok                                                           OK
global_button_cancel                                         cancel                                                       Cancel
global_button_clean                                          clean                                                        Clean
global_button_remove                                         browser.collection.remove                                    Remove
global_button_remove                                         remove                                                       Remove
global_button_add                                            add                                                          Add
global_button_edit                                           edit                                                         Edit
global_button_edit                                           edit.menu                                                    Edit
global_spinner_select_option                                 track.instrument.default-select-option                       -- Select --
global_spinner_select_option                                 tuning.value.select                                          -- Select --
action_menu_file                                             browser.menu.file                                            File
action_menu_file                                             file                                                         File
action_menu_edit                                             edit                                                         Edit
action_menu_edit                                             edit.menu                                                    Edit
action_menu_view                                             view                                                         View
action_menu_transport                                        transport                                                    Player
action_menu_caret                                            UNKNOWN                                                      Caret
...
```

If no match was found, the second column will read "UNKNOWN".

Short explanation how it works (-h):

```
#
# The script uses the following logic to determine the corresponding stings:
# Same english text => Entries mean the same, e.g.:
#   messages.properties:   key_desktop=english bla bla
#   strings.xml:           <string name="key_android">english bla bla</string>
#   => val_desktop = val_android = "english bla bla"
#   => val_desktop means the same as val_android => take over the content of val_desktop to val_android for all languages
#
```

Of course, if the variable names of the desktop and Android versions were the same, it would be much easier to populate the Android files from the desktop version. Another question is how to handle messages that are only present in the Android version; you could put them in the desktop files as android_only_* for example. But also the desktop version has many unused and duplicate entries, another TODO.... So I decided to write this script and maybe think about a clean solution later.

guiv42:
OK, thank you for all these details. Just like you say: another TODO...

Issue #92, opened 2023-08-21
Stop and resume playback from original spot


alejandro6626:
Up to version 2023-05-27 when stopping playback (e.g. using spacebar) the active measure or cursor goes back to the spot from where playback was started, meaning that if you hit play or press spacebar again, playback will start over again from the same place, unless you manually change the current measure or starting point. This is extremely useful for practicing passages over and over again (just hitting spacebar) and was always a good feature of tuxguitar that has always existed. However, from that version onwards, when you hit stop and then you resume, playback will continue from where it left, meaning that if you want to practice a section again, you need to manually look for the measure where it starts which is not practical for practicing, specially if you're standing with a guitar hanging from your shoulder.
I think that the old behaviour should be restored or that at least it can be selected from preferences.

guiv42:
We're entering a subjective area here: this new behavior was implemented following users' requests. See [here](https://sourceforge.net/p/tuxguitar/discussion/522984/thread/65c8acf27b/) and [here](https://sourceforge.net/p/tuxguitar/feature-requests/101/).
I also considered old behavior inconsistent: it wasn't logical to have "pause" and "stop" buttons behaving identically.
Note: this new behavior is consistent with Android app.

That being said, your use case is relevant, and there might be a way to address it also. A possible solution could be to create a new action "play/stop", accessible through a shortcut. 
This way one could allocate spacebar either to play/pause or to play/stop actions.

What do you think?

note: not yet evaluated the complexity of such a feature

alejandro6626:
Yes, it's subjective. However I noticed this change immediately as I've used Tuxguitar for a very long time and I could say the behavior has been always like that as far as I remember.

In my case, I use Tuxguitar as a backing track to practice guitar solos, so I need to stop and resume from the same spot all the time, thousand of times, and doing that with the spacebar was really helpful.

Assigning a shortcut key to the stop action works, e.g. "c", i just need to start play mode with spacebar and then stop with "c". It could take a while for me to get used to it, as I probably will instintively hit spacebar to stop at the beginning.

As you say, space bar is assigned to play button, and the play button turns into pause, so spacebar is truly assigned to "play/pause" and not "play/stop", which doesn't exist. (Actually in the shortcuts Window, the instead of "play" action it should say "play/pause"). If you create a play/stop action, then I could move the spacebar away from play/pause and assign it to the play/stop and everyone would be happy.

[...]

PR #94, opened 2023-08-21
add Play/Stop action

guiv42:
see #92 

note: logically, "action.transport.play" could have been renamed into "action.transport.playPause"
but it would break space bar shortcut definitions in users configuration... (for users already using this repo).


guiv42:
not ready for a merge yet, forgot link with plugins...

guiv42:
better now

PR #136, opened 2023-10-10
fix CVE-2020-14940

guiv42:
see:
http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2020-14940
https://sourceforge.net/p/tuxguitar/bugs/126/
https://bugzilla.opensuse.org/show_bug.cgi?id=1173633
https://logicaltrust.net/blog/2020/06/tuxguitar.html https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html

Issue could be reproduced on Linux before the fix, as described by sourceforge page listed above
note: needed to de-activate firewall to reproduce issue. Not all TuxGuitar files mentioned in this page have been modified, as some of them do not parse input xml files. Then they should not be concerned by vulnerability:
- TuxGuitar-musicxml/src/org/herac/tuxguitar/io/musicxml/MusicXMLWriter.java
- TuxGuitar/src/org/herac/tuxguitar/app/system/keybindings/xml/KeyBindingWriter.java
- TuxGuitar/src/org/herac/tuxguitar/app/tools/browser/xml/TGBrowserWriter.java

note: protection does not seem to be supported on Android, so just try to activate it, and ignore if it fails (or else Android version can no more open .gp and .gpx files)
basically: this patch doesn't provide full coverage

also (independent from CVE):
GPXDocumentReader: removed warnings from deprecated Integer constructors

### DIFF
--- a/TuxGuitar-community/src/org/herac/tuxguitar/community/browser/TGBrowserResponse.java
+++ b/TuxGuitar-community/src/org/herac/tuxguitar/community/browser/TGBrowserResponse.java
@@ -35,6 +35,12 @@ public class TGBrowserResponse {
 	
 	private void initialize(InputStream stream) throws Throwable {
 		DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+		// CVE-2020-14940
+		try {
+			factory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+			factory.setXIncludeAware(false);
+		} catch (Throwable throwable) {
+		}
 		DocumentBuilder builder = factory.newDocumentBuilder();
 		this.document = builder.parse(stream);
 	}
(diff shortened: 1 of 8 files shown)


helge17:
Thanks for the patch! I was able to reproduce the POC from the Sourceforge page on Linux, but not on Android (with Python in Termux, TuxGuitar without and with this PR).

When I open on Android, poc1.gp, TuxGuitar shows the error: Attempt to invoke interface method 'int org.w3c.dom.NodeList.getLength()' on a null object reference

Opening poc2.gp shows: Invalid file format

In both cases, no request is logged in the web server (the server is running correctly, I can access the files at http://127.0.0.1:8000/ with the local browser, and the access is logged by the web server).

guiv42:
Thanks for the test on Android, that's good news :)
I'll post a message on sourceforge to thank the author of that post: clear description of the issue, procedure to reproduce (with attached files), procedure to patch, really this was a fantastic job

note: I noticed this issue while looking at openSuse bug tracker. OpenSuse maintainer for TuxGuitar did a fantastic job following my request to update to 1.6.0, so I thought it would be fair to have a look at their bug tracker ;)

PR #139, opened 2023-10-21
Rework of cursor position vs. player

guiv42:
Evolution is quite complex, so I would appreciate some complementary tests if possible.
I tried to limit the risk by implementing automated tests, but I would not be surprised if a few bugs remain.
Note: I could not fully test with Jack plugin enabled.

# Description of the evolution
Where does the player start when hitting *play*? Where does the cursor go when hitting *pause* or *stop*?

Objective of this evolution is to rework how cursor and player interact, especially considering the "click & drag" feature introduced in 1.6.0. Examples of scenarios:
- select measures 4 to 6, type F9, click "play looped", click OK then *play*: it should loop from 4 to 6 (starts playing 6 before this evolution)
- after selecting something with click&drag, *play* should play once the selected area (plays from cursor to end of song before this evolution)

# User interface
From user point of view, the new behavior is the following:

- Priority to "click&drag" selection: after selecting something, *play* action plays this selection, just once and independently from player mode settings (F9). E.g. if a loop is defined in player mode it is ignored.
- Generally, hitting *play* starts playing where the cursor is. Exception: if something was just selected, it starts playing at beginning of selection, independently from cursor position
- When hitting *pause*, cursor is moved to the measure where *pause* was hit
- When hitting *stop* or when end of song is reached, in absence of selection cursor is moved to the measure where *play* was hit. With a selection active, cursor comes back to beginning of selection
- hitting *stop* while player is not playing does nothing.

# Implementation
From software point of view, there are 2 different indexes:

- the cursor (named `Caret` in the code), that user can move with keyboard, mouse, toolbar buttons, menu items...
- a `tickPosition`, internal to `MidiPlayer`, moving when song is played

I tried to de-correlate movements of `caret` and `tickPosition`:

- whatever user does when player is not playing, `tickPosition` is no more updated
- when player plays, `caret` position remains unchanged

The 2 indexes are aligned when player changes state:

- when player starts to play, `tickPosition` is updated considering selection state, `caret` position and player mode (e.g. loop configured)
- when player stops, `caret` is moved considering user action (*stop* or *pause*), current `tickPosition` and player starting point

# Limitations

With this evolution, player always plays **full** measures.
E.g. if selection covers last beat of measure 4 to first beat of measure 5, then *play* action plays full measures 4 & 5. Similarly, with one single beat selected a full measure is played.
This could be reworked in a second step (significant complexity, impacting several implementations of `MidiSequencerImpl` class).
A test corresponding to this use case has been added in `TGTestPlayerPosition`, it is the only failing test.

Automated tests may not be 100% reliable, there could be false positives (not too often hopefully). I did not succeed to implement an automated test for the case where a "Close Repeat" is present in the selected area. This case was only tested manually.

### DIFF
--- a/TuxGuitar-jack/src/org/herac/tuxguitar/jack/sequencer/JackMidiPlayerStarter.java
+++ b/TuxGuitar-jack/src/org/herac/tuxguitar/jack/sequencer/JackMidiPlayerStarter.java
@@ -31,7 +31,7 @@ public class JackMidiPlayerStarter implements TGEventListener {
 	}
 	
 	public void start(){
-		TGTransport.getInstance(this.context).playPause();
+		TGTransport.getInstance(this.context).playPause(null, null);
 	}
 	
 	public void processCountDownStarted() {
--- a/TuxGuitar-lib/src/org/herac/tuxguitar/player/base/MidiPlayer.java
+++ b/TuxGuitar-lib/src/org/herac/tuxguitar/player/base/MidiPlayer.java
@@ -72,6 +72,10 @@ public class MidiPlayer{
 	
 	private int loopEHeader;
 	
+	private TGBeat selectionStartBeat;
+	private TGBeat selectionEndBeat;
+	private boolean isNewSelection;
+	
 	private long loopSPosition;
 	
 	private boolean anySolo;
@@ -107,7 +111,11 @@ public class MidiPlayer{
 			this.lock();
 			
 			this.stop();
-			this.tickPosition = TGDuration.QUARTER_TIME;
+			if (this.selectionStartBeat!=null) {
+				this.tickPosition = this.selectionStartBeat.getStart();
+			} else {
+				this.tickPosition = TGDuration.QUARTER_TIME;
+			}
 			this.setChangeTickPosition(false);
 		} finally {
 			this.unlock();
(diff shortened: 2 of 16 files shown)


helge17:
The player behaves very well now, thanks for the improvement!

I just found one very small incosistency: Suppose you have an Open Repeat in measure 5 and a Close Repeat in measure 6 (one repetition). If you set the player to loop from measure 4 to measure 5 it plays 4-5->4-5->4-5...; I would expect it to play 4-5-5->4-5-5... If you loop over measures 4 to 6 it plays 4-5-6-5-6->4-5-6-5-6..., as expected.

guiv42:
Thanks for your feedback!
About this specific scenario: this is the way TG behaved before the evolution in case a loop is selected. With *Open Repeat* in 5 + *Close Repeat* in 6 + loop defined on interval [4-5] it plays 4-5-4-5. And my interpretation is that it makes sense.

Just looking at the selected interval [4-5], an *Open Repeat* is present... but there is no guarantee that a *Close Repeat* will be present later! That would be a little stupid, but it's possible, and in this case there would be no reason to play 4-5-5.
Another scenario (still stupid but still possible): *Open Repeat* in the selected/loop interval, but with another *Open repeat* before the first *Close Repeat* is found. E.g. *Open Repeat* in 5, *Open Repeat* in 7, *Close Repeat* in 8, then select (or loop) over interval [4-5]. Again, no reason to play 4-5-5.

In other words, I would prefer to keep this behavior unchanged.
When a selection/loop is active, analysis of the content of this interval should be sufficient to decide what to play. If the analysis shall be done on the whole song this add a very significant level of complexity to handle all possible cases (even if stupid)

helge17:
Thanks for the explanation, that makes sense to me.

Issue #141, opened 2023-10-28
nice to have zoom (in, out and reset) in main toolbar


sugizo:
nice to have 
zoom button 
and sub button for zoom in, zoom out and zoom reset
in main toolbar

thanks and best regards

guiv42:
Hi
the easiest way to zoom in/out is currently (Ctrl+mouse wheel) or (Ctrl+2fingers touchpad). From pre-release 2023-10-26 it is also possible to configure keyboard shortcuts.

From code perspective, adding buttons to the toolbar is fairly simple, it's already been done a few years ago on [SF](https://sourceforge.net/p/tuxguitar/patches/34/) (but not merged).
It seems to me the most complex part is to update the icons. Icons for "zoom in/out" can be found [here](https://github.com/brockdev/tuxguitar/tree/master/TuxGuitar/share/skins) (following link in SF post above), but:
- there is no icon for "zoom reset"
- icons are only available in png format, no vector graphics
- some work is needed to adapt to all skins (currently 9 skins are available)

If someone could provide some help for the icons, I can update the code.

helge17:
@guiv42, if you update the code I'll care about the icons.

sugizo:
why not adapt it like power tab, 
they don't use icon, 
but use percentage (%) instead, 
so user can type 
or 
scroll the percentage value they chosen

e.g. zoom = 125 % on powertab
![image](https://github.com/helge17/tuxguitar/assets/2418295/e875d77a-7a75-42be-b64f-d2d60db28080)

guiv42:
I like the idea, but from software point of view it's much more complex. I already had a try a couple of months ago, and I finally abandoned (or at least postponed!). The software structure of the toolbar makes it easy to integrate buttons, but a significant rework is needed to integrate other widgets, like the one suggested above.
So I prefer to do it in 2 steps: simple adaptation first, and one day a complete rework of the toolbar (that might be an opportunity to also integrate player dialog in toolbar)

helge17:
Thanks for the patch! I'll care about the Symbolic & Breeze Icons, previews etc. I think I will also replace the other zoom icons with the "original" ones from https://invent.kde.org/frameworks/oxygen-icons5/-/tree/master/22x22/actions/ But give me a few days, too much to do currently...

guiv42:
Thanks
For sure this graphical edition task is really time consuming. But take your time, nothing's urgent.

sugizo:
[...]

PR #146, opened 2023-10-30
add zoom controls in main toolbar

guiv42:
see #141 
limitations
- icons only in png format, no vector graphics
- identical icons for all skins


guiv42:
I just noticed I forgot to update the screenshots in the help section

PR #147, opened 2023-10-30
add a (basic) text filter in KeyBindingEditor

guiv42:
many actions are available there
ability to search in list should be useful


guiv42:
I also forgot to update the help section...

guiv42:
That's better

Issue #153, opened 2023-11-17
Help wont open in Browser on Ubuntu System


BeFrankly:
Hello,

first I would say that it is great, that someone took the code and goes on with developing and fixing Tuxguitar. 

Yesterday I installed version 1.6.0. via dpkg on my Ubuntu 22.04.3 LTS and needed some help/documentation. So from the upper menu of tuxguitar i opened the help and so the standard browser, Firefox in my case, opened. The URL was file:///opt/tuxguitar/share/help/index.html and browser tells "error: file not found.".

In a terminal i did ls -l  /opt/tuxguitar/share/help and all the files are there and readable by owner, group and everyone. So why does it not open. Same with Chromium browser.

guiv42:
This could be a limitation of Firefox, which version are you using?
See [there](https://askubuntu.com/questions/1435757/firefox-cant-open-files)

BeFrankly:
Oh dear. I really need to get to grips with the configuration of Snap. Ok, security is an important point. 
I'll give it a try and report back if it works. Might take a while because I need to practise a few pieces on the guitar first. :-)

BeFrankly:
Dear @guiv42 , I have tried the solution with apparmor from the link you provided. It does not work. But let's think again:
It is not a good idea to break Snap's security wall by editing the apparmor configuration just to see a documentation. In my opinion, it's better to provide a different path and maybe a different format of tuxguitar documentation that works with snap or other security restrictions.

guiv42:
Hi @BeFrankly 
In fact I did not suggest to change this configuration, objective of this link was just to highlight that other Ubuntu users had similar issues. I just wanted to know if this could be the cause of your problem (I don't know either Ubuntu or Snap). Sorry if it was not clear. From your feedback I understand the problem is effectively there.
Anyway, the complete Ubuntu community will face this issue, so I agree it would be useful to find a solution.

Writing and maintaining doc is really time-consuming, and @helge17 already completely re-wrote this doc for 1.6.0. That would be for sure a significant amount of work to re-write it again. And static html format has the advantage to be easily deployed cross-platform (well... at least theoretically!). So I must admit I'm not enthusiastic about an evolution of format. Any idea about another possible format?

Another option could be to make this html content available on a web site (possible also to add a menu item "online help" targeting this site). 
But I would find this really frustrating. As you said "OK, security is an important point", so... forcing user to connect to the Internet to fetch some content already available locally... I'm not convinced it really increases security!

[...]

PR #159, opened 2023-11-26
display doc in embedded browser

guiv42:
to remove dependency to external web browser
Specific use case for Ubuntu, where distro's web browser cannot access local resources
see #153

Limitations:
- pom files updated for all configurations, but only tested with Linux/JFX, Linux/SWT, Win10/SWT
- no embedded controls for navigation (e.g. back, forward)

The method "openUrl" of application object is now unused, but I did not remove it: it can be used in future versions to link to an external website for online help

### DIFF
--- a/TuxGuitar-ui-toolkit-jfx/pom.xml
+++ b/TuxGuitar-ui-toolkit-jfx/pom.xml
@@ -45,6 +45,11 @@
 		</dependency>
 		<dependency>
 			<groupId>org.openjfx</groupId>
+			<artifactId>javafx-web</artifactId>
+			<version>${javafx.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.openjfx</groupId>
 			<artifactId>javafx-controls</artifactId>
 			<version>${javafx.version}</version>
 		</dependency>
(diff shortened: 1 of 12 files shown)


helge17:
Linux SWT, help not starting; error:
org.herac.tuxguitar.action.TGActionException: No more handles because there is no underlying browser available.
Please ensure that WebKit with its GTK 3.x bindings is installed (WebKit2 API level is preferred). Additionally, please note that GTK4 does not currently have Browser support.

FreeBSD JFX: Does not build.

Probably some dependency problems, I have to investigate...

guiv42:
Linux SWT: I have "libwebkit2gtk-4_0-37" installed on my system, this might be the required dependency?

helge17:
Indeed, libwebkit2gtk-4.0-37 is needed. I only had libwebkit2gtk-4.1-0 install, this does not work.

guiv42:
Okay, so at the end it boils down to a dependency issue: where do we expect to find a working web engine?
- before PR: constraint exported to the OS, fails in Ubuntu context
- after PR: manage constraint internally, fails in some other situations

I don't know exactly how to handle this. At least for all runtime errors, a simple modification could be implemented when user types F1:
- try to display doc in embedded browser
- if it fails, catch exception and fallback to historical behavior (start external web browser).

For Linux, this could end up in a "recommended" package in deb/rpm package.

For build-time errors (FreeBSD)... I have no idea

guiv42:
> Indeed, libwebkit2gtk-4.0-37 is needed. I only had libwebkit2gtk-4.1-0 install, this does not work.

If I try to uninstall it, my package manager tells me it also needs to uninstall TuxGuitar 1.6.0 (OpenSuse rpm version). In fact, TuxGuitar 1.6.0 rpm has "eclipse sw >= 4.13" as an explicit dependency. And the version of eclipse-swt installed in my system (4.15-11.2) has an explicit dependency to libwebkit2gtk-4.0.so.37...

PR #161, opened 2023-11-28
restore FreeBSD/JFX build

guiv42:
see #159 (especially [freebsd/jfx issue](https://github.com/helge17/tuxguitar/pull/159#issuecomment-1826788310)), and #153

I could not run this exact build script in my environment, but I performed the same steps manually. Then, build of FreeBSD/JavaFX version succeeded.
So, I recommend to check before merging.

note: I could not either test if the produced app works correctly (I just configured build environment in freebsd, but no x server yet)


helge17:
Thanks! I was just working on the same problem.

guiv42:
It's fine if I could fix the regression I had introduced ;)
(and it was also a good opportunity for me to - finally - set up a FreeBSD environment)

PR #180, opened 2023-12-26
Just deleted TGMain class because it is not needed.

BeFrankly:
So now we have just one main method in the codebase.


guiv42:
Without this class, Android & Windows & Linux & FreeBSD versions build correctly. Linux & Android versions work correctly.
Not tested in macOS configuration.
All start scripts in all configs seem to target TGMainSingleton, and Eclipse does not find any dependency.
So, it's OK for me.
 
Note: I don't have the history of this file, and no explanation of the redundancy of "main()" with TGMainSingleton.

PR #210, opened 2024-01-16
Update tunings.xml

belfie13:
fixed A# 6-string guitar tuning and added standard tunings for 4-string bass


guiv42:
thanks for contributing
Just for information, in latest snapshots you can add and save your own custom tunings

belfie13:
yeah i just saw that :D kool..

belfie13:
i haven't done much contributing so don't be shy to point out anything, the added presets were probably overboard 😬 👍

Issue #211, opened 2024-01-16
importing midi type 0 can lead to negative fret numbers


guiv42:
tested on 1.6.0, Linux openSuse Tumbleweed.

Importing a type 0 midi file from archive present in [this post](https://github.com/helge17/tuxguitar/issues/205#issuecomment-1893981907) leads to negative fret numbers displayed.
This file contains several instruments in one single midi track. When TuxGuitar imports file, it merges all instruments notes into one single track. So, in the last beat, more than 6 simultaneous notes are present, with several instances of one single note. Most probably this situation creates the issue when TuxGuitar tries to find a string to play each note

belfie13:
have a look at [CIID-Heaven.Hell-BibleBlackBass2.mid.txt](https://github.com/helge17/tuxguitar/files/13950044/CIID-Heaven.Hell-BibleBlackBass2.mid.txt) from issue #205 which has negative numbers (example at bar 26) and it is a type 1 midi file (should be, i can't verify that). also the notes used different channels on the one track in Logic Pro before exporting to midi, this could be an issue.

just fyi, the channels in logic can be used to determine which string the note should be played on (or tracks can be configured so the channels represent how high on the fretboard / the lowest position to play the note)

guiv42:
> have a look at [CIID-Heaven.Hell-BibleBlackBass2.mid.txt](https://github.com/helge17/tuxguitar/files/13950044/CIID-Heaven.Hell-BibleBlackBass2.mid.txt) from issue https://github.com/helge17/tuxguitar/issues/205

I thing you mean #209, yes I need to have a look at that. It might be another problem though.
when I import CIID-Heaven.Hell-BibleBlackBass2.mid file with -12 transposition, and with latest snapshot, I cannot see any negative number. So my assumption currently is that #209 is more a problem of wrong tuning selection, to be confirmed when I find some time to analyze

guiv42:
After a short analysis of the code, it seems problem can also occur with a type 1 midi, even without any tuning issue.
Distribution of notes to strings can be improved:
- it processes notes in one beat following order of midi events. It seems to me processing from lowest to highest should increase chances of success (globally less possible places on a fretboard for lowest notes)
- it seems it assumes all notes *can* be played, which can be false, if source midi file was not written for guitar. In some cases, unplayable notes *shall* be discarded

analysis still ongoing, but it seems it can be fixed.

belfie13:
[...]

Issue #209, opened 2024-01-16
Changing number of strings does not function as expected


belfie13:
background: When I imported a midi (exported from logic pro x) I had many issues which could be related to the midi file and not tux guitar. some notes had negative numbers which corresponded to the right notes but could not be edited (ie move to higher/lower strings or change note value) and had to be deleted before i could enter the proper note.

one thing that I could not find a way to work around was when changing tunings (ie the bass track loaded as a 6 string guitar), notes dissappeared. I'll explain with pictures

watch the notes on the bottom string as i try and delete the top string from the instrument.

note: i believe i unchecked transpose affected notes in this example but the results were the same.

![Screen Shot 2024-01-16 at 12 51 02](https://github.com/helge17/tuxguitar/assets/39270867/dc521ee0-2137-41ca-8612-0fddc8ff6c19)
![Screen Shot 2024-01-16 at 12 51 18](https://github.com/helge17/tuxguitar/assets/39270867/fffe45d1-3a17-41eb-8464-eda224569e89)
![Screen Shot 2024-01-16 at 12 52 40](https://github.com/helge17/tuxguitar/assets/39270867/289fad4d-e5b6-47ff-8788-f11f765922f8)

belfie13:
er should have used an example with more notes but you can see the effect on bar 26

guiv42:
Are your tracks in the original midi file adapted to the instruments (notes height I mean)? TuxGuitar chooses tuning and number of strings considering lowest/highest notes of each track. If TuxGuitar selected a 6-string guitar for your bass track, then notes in this track can be played by a guitar. A possibility is everything is an octave too high (I don't know about bass, but guitar treble clef is implicitly 1 octave below standard clef).
Another user had similar issues because of too high notes, so TuxGuitar did not choose an appropriate tuning (see #150). 

Which version are you using? 1.6.0 I suppose ?
A few bugs were fixed after 1.6.0 about midi import (e.g. negative numbers).

Could you please check with the latest snapshot? If problem persists, try to directly transpose (-12) when importing midi.
If problem still persists, could you please attach here a small midi file triggering the issue?

belfie13:
still happens with tuxguitar-2024-01-13-snapshot-macosx-swt-cocoa-x86_64 
(i tried jfx version but the menu didn't work)

i tried editing the number of strings by deleting the top one and by choosing a preset tuning.

[...]

PR #215, opened 2024-01-18
Fix negative frets when importing midi, update help

guiv42:
See #211, also mentioned in #209

### DIFF
--- a/common/TuxGuitar-lib/src/org/herac/tuxguitar/song/models/TGNote.java
+++ b/common/TuxGuitar-lib/src/org/herac/tuxguitar/song/models/TGNote.java
@@ -13,7 +13,7 @@ import org.herac.tuxguitar.song.factory.TGFactory;
  * 
  * TODO To change the template for this generated type comment go to Window - Preferences - Java - Code Style - Code Templates
  */
-public abstract class TGNote {
+public abstract class TGNote implements Comparable<TGNote>{
 	private int value;
 	private int velocity;
 	private int string;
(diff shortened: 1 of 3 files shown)


belfie13:
Is the playback directly related to whats visible in the score? i mean logic still plays the notes that it cant display on the tab (which arent possible to play given the instrument configured). they will be visible in the score, otherwise if you just throw away half an import then there's no way to recover, might just simply be a matter of changing the tuning and it can be played.

guiv42:
Wow you're asking for complex things!
No, that's not the way I did it. Keeping in mind TuxGuitar is essentially a *tablature* editor, and not a *MIDI* editor, I preferred to do it in a simpler way: when importing a chord which cannot fit in the fretboard, just discard some notes.
Whenever you try to introduce in TuxGuitar some song which is not adapted for fretted strings instruments, you'll get a lot of trouble and side effects. On my side, I would definitely not understand having "hidden" notes.
I completed the ["help"](https://github.com/guiv42/tuxguitar/blob/master/desktop/TuxGuitar/share/help/file_formats.html) section of app about this, don't hesitate to review it (and correct if if needed). It's static html, so just download the file and open it with a web browser, see "MIDI file import" section.

Important note: please consider this fix is **only related to midi import** feature. So this is completely independent from processing involved when you change tuning or instrument (so #209).

I don't know if you can build from source, if not just wait for the next pre-release to be published. This fix plus "import/transpose -12" should be a good starting point to go further in #209.

PR #200, opened 2024-01-13
Fix French translation in tuxguitar.desktop

trebmuh:
HTH

PR #247, opened 2024-02-18
French translation update and (almost) completion

trebmuh:
Following #200


guiv42:
Thanks!
I'm just a little puzzled by the "piano à queue électrique" (note that "grand piano électrique" was not really better).
I'm not a drummer, so this may not be relevant:  I'm also surprised not to find anywhere "rim shot" (even in the english version), there must be another way to name it that I don't know...

another thing: it seems that when translated string is identical to the english version, usual way to do it was to leave field empty. I don't know why, but I prefer the way you have done it. Even if the French string is identical to the English one, at least it makes explicit that string *is* translated. So, fine to keep it like this.

trebmuh:
Thanks for the merge!

"grand piano" (english) translates to "piano à queue" (french). To convince yourself, you can see that on WP : https://fr.wikipedia.org/wiki/Piano_%C3%A0_queue where you can use the dropdown language menu à the top.

As far as I know, "rim shot" is used as is in French. Never heard another way calling it I think.

> another thing: it seems that when translated string is identical to the english version, usual way to do it was to leave field empty. I don't know why, but I prefer the way you have done it. Even if the French string is identical to the English one, at least it makes explicit that string is translated. So, fine to keep it like this.

Perfect, exactly the same for me. It makes it clear for everyone that it is translated.

Just as a wish list for the future: it would be easier for translation/translators to use something like gettext as a translation system. I understand that it might take a lot of development workforce for nothing "new" for the end user at the end of the day. So no pressure for me, just it would make translators life easier.

helge17:
>> another thing: it seems that when translated string is identical to the english version, usual way to do it was to leave field empty. I don't know why, but I prefer the way you have done it. Even if the French string is identical to the English one, at least it makes explicit that string is translated. So, fine to keep it like this.

> Perfect, exactly the same for me. It makes it clear for everyone that it is translated.

The misc/messages_desktop_sort.pl script replaces untranslated messages or messages that are exactly the same as in messages.properties with "# message.key=". I often used the script to clean up the mess in the messages_*.properties files after starting this fork.
For a non-native speaker it is quite difficult to recognize if "message.key=english value" means that the English term is used in this language or if it simply means that nobody has translated this message yet. I have therefore decided to comment out these entries without a value. One advantage is that if the English message changes, this change is automatically transferred to all other languages. To me, this seemed to be the best solution, but of course this is open to debate.

guiv42:
For a newcomer, it seems to me easier to make explicit what *is* translated, even if the translated text is the same as the original. First time I updated the French translation I started to update these fields (identical to English) until I realized that it was intentional.
Anyway it's not a big deal. If we really want to ease the job of translators, it could be valuable to switch to a more user-friendly translation system. Either "gettext" as mentioned above, or weblate.org, or... I don't know.

PR #259, opened 2024-03-03
split README in 2 independent files

guiv42:
@helge17: I will not merge this PR myself, I would like you to review it first.

My objective is mainly to ease access to "contribute.md" from repo's main page (link was deleted at my request, because of markdown syntax appearing in help/about dialog: 566b3c5b5ae5a1de4430e20b477dc3746d9fad00)
Not very comfortable to split "README" in 2, especially for future maintenance. Not ideal also to embed in deskop app an absolute link to "contribute.md".
Don't hesitate to reject PR if you can think of a better solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,12 +22,10 @@ TuxGuitar includes the following third party products:
 
 ## Installation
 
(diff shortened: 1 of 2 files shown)


helge17:
Thanks, I don't have a better idea either.

PR #276, opened 2024-03-12
Added guideline to enable run/debug with plugins

profOnno:
I think this pull request needs some testing. It works for me, not sure it works for the majority.
Notice I used the strike through in the `important note`.

### DIFF
--- a/doc/contribute.md
+++ b/doc/contribute.md
@@ -98,10 +98,13 @@ This walkthrough assumes that you will be using the Eclipse Development Environm
 Eclipse supports many processes around software projects, even writing documentation in markdown.  For information about downloading and installing it see <https://eclipse.org>.
 Our examples and screenshots here are from Eclipse Version 2023-12.
 
-Important note: this section still needs to be completed. Known limitations:
+*Important note*: this section still needs to be completed. Known limitations:
 
 - this procedure does not enable to build the *native modules* in Eclipse. Therefore, before sending a pull request it is highly recommended to build the application as defined in [install.md](../INSTALL.md), that is with `-P native-modules` option active.
-- this procedure does not enable to include TuxGuitar plugins in debug configuration
+- ~~this procedure does not enable to include TuxGuitar plugins in debug configuration~~
+
+To get the enable the plugins in debug configuration make sure to set the `Working directory` in the `Run` configuration, `Arguments` tab to `other`, "${workspace_loc}/git/tuxguitar/desktop/build-scripts/tuxguitar-linux-swt-x86_64/target/tuxguitar-SNAPSHOT-linux-swt-x86_64". To make debugging work, add "gm-settings-SNAPSHOT.jar" and "gm-utils-SNAPSHOT.jar to the `dependencies` tab of the `Run` configuration.
+
 
 #### Fork the origin tuxguitar repository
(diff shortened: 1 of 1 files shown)


doc/contribute.md
```diff
@@ -116,7 +119,7 @@ In the next step you will probably only **see the master branch. It has to been
 
 <img src="images/contribute/eclipse-git-clone-step2.png" width="50%" alt="Git Clone Step 2, select branches">
 
-As a last step you have to choose where on your local drive the repo should be placed. In my case, I created a folder named "git" in my Eclipse workspace.
+As a last step you have to choose where on your local drive the repo should be placed. In my case, I created a folder named <a name="git_folder_name"></a> "git" in my Eclipse workspace.
```
guiv42, line 122: I don't understand this modification: the content of <a> tag is empty, so nothing is displayed

doc/contribute.md
```diff
@@ -98,10 +98,13 @@ This walkthrough assumes that you will be using the Eclipse Development Environm
...
 
-Important note: this section still needs to be completed. Known limitations:
+*Important note*: this section still needs to be completed. Known limitations:
 
 - this procedure does not enable to build the *native modules* in Eclipse. Therefore, before sending a pull request it is highly recommended to build the application as defined in [install.md](../INSTALL.md), that is with `-P native-modules` option active.
-- this procedure does not enable to include TuxGuitar plugins in debug configuration
+- ~~this procedure does not enable to include TuxGuitar plugins in debug configuration~~
+
+To get the enable the plugins in debug configuration make sure to set the `Working directory` in the `Run` configuration, `Arguments` tab to `other`, "${workspace_loc}/git/tuxguitar/desktop/build-scripts/tuxguitar-linux-swt-x86_64/target/tuxguitar-SNAPSHOT-linux-swt-x86_64". To make debugging work, add "gm-settings-SNAPSHOT.jar" and "gm-utils-SNAPSHOT.jar to the `dependencies` tab of the `Run` configuration.
+
```
guiv42, line 107: unfortunately it does not work, at least in my configuration :(

guiv42 (commented):
I tested the updated procedure, and in my configuration it does not work (I still don't have access to the plugins while debugging). By the way, I do not see any direct link between gm-settings, gm-utils, and the activation of plugins (?).

I need to spend some more time on this debugging issue.
As is, this PR cannot be merged, sorry :(

doc/contribute.md
profOnno: It is an anchor, later in the document there is a link to jump to exactly this location.

doc/contribute.md
profOnno: I'll look in to this some more. For me the plugin, I tested was `desktop/tuxguitar-musicxml`. So, I hope, our definition of **plugin** are the same.

profOnno:
> I do not see any direct link between gm-settings, gm-utils, and the activation of plugins (?).

I got an error, it could not find it. I presumed (shouldn't do that), it had something to do with the plugin loading.
I'll try to find a `path`, so it will find al jars if it needs one. This is only of value, when it works at all, and we're talking about same thing (plugin wise).

doc/contribute.md
guiv42: OK, clear

profOnno:
Some more info:

This is a screenshot of my run configuration.
my argument settings
Notice, I'm not using `git/tuxguitar` as base.

The result with these settings:
run_arguments_other_result

The result without these settings:
run_arguments_default_result

This for debug to work another step is needed.
But thirst make sure we're on the same page, on this issue (if it is one :laughing:).

Is this the same on your end, or do you get already get a full list of plugins?

guiv42:
PR closed, superseded by #359

PR #277, opened 2024-03-12
Get run/debug working.

profOnno:
Changed working directory in run configuration.
Added tuxguitar-gm-settings-SNAPSHOT.jar and
tuxguitar-utils-SNAPSHOT.jar to classpath in run configuration.

### DIFF
--- a/development/Tuxguitar SWT.launch
+++ b/development/Tuxguitar SWT.launch
@@ -17,6 +17,8 @@
         <listEntry value="&lt;?xml version=&quot;1.0&quot; encoding=&quot;UTF-8&quot; standalone=&quot;no&quot;?&gt;&#10;&lt;runtimeClasspathEntry containerPath=&quot;org.eclipse.m2e.MAVEN2_CLASSPATH_CONTAINER&quot; path=&quot;5&quot; type=&quot;4&quot;/&gt;&#10;"/>
         <listEntry value="&lt;?xml version=&quot;1.0&quot; encoding=&quot;UTF-8&quot; standalone=&quot;no&quot;?&gt;&#10;&lt;runtimeClasspathEntry id=&quot;org.eclipse.jdt.launching.classpathentry.variableClasspathEntry&quot;&gt;&#10;    &lt;memento path=&quot;5&quot; variableString=&quot;${workspace_loc}/externals/swt-4.13-gtk-linux-x86_64/swt.jar&quot;/&gt;&#10;&lt;/runtimeClasspathEntry&gt;&#10;"/>
         <listEntry value="&lt;?xml version=&quot;1.0&quot; encoding=&quot;UTF-8&quot; standalone=&quot;no&quot;?&gt;&#10;&lt;runtimeClasspathEntry path=&quot;5&quot; projectName=&quot;tuxguitar-ui-toolkit-swt&quot; type=&quot;1&quot;/&gt;&#10;"/>
+        <listEntry value="&lt;?xml version=&quot;1.0&quot; encoding=&quot;UTF-8&quot; standalone=&quot;no&quot;?&gt;&#10;&lt;runtimeClasspathEntry internalArchive=&quot;/tuxguitar-gm-settings/target/tuxguitar-gm-settings-SNAPSHOT.jar&quot; path=&quot;5&quot; type=&quot;2&quot;/&gt;&#10;"/>
+        <listEntry value="&lt;?xml version=&quot;1.0&quot; encoding=&quot;UTF-8&quot; standalone=&quot;no&quot;?&gt;&#10;&lt;runtimeClasspathEntry internalArchive=&quot;/tuxguitar-gm-utils/target/tuxguitar-gm-utils-SNAPSHOT.jar&quot; path=&quot;5&quot; type=&quot;2&quot;/&gt;&#10;"/>
     </listAttribute>
     <stringAttribute key="org.eclipse.jdt.launching.CLASSPATH_PROVIDER" value="org.eclipse.m2e.launchconfig.classpathProvider"/>
     <booleanAttribute key="org.eclipse.jdt.launching.DEFAULT_CLASSPATH" value="false"/>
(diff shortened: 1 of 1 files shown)


development/Tuxguitar SWT.launch
```diff
@@ -17,6 +17,8 @@
...
         <listEntry value="&lt;?xml version=&quot;1.0&quot; encoding=&quot;UTF-8&quot; standalone=&quot;no&quot;?&gt;&#10;&lt;runtimeClasspathEntry path=&quot;5&quot; projectName=&quot;tuxguitar-ui-toolkit-swt&quot; type=&quot;1&quot;/&gt;&#10;"/>
+        <listEntry value="&lt;?xml version=&quot;1.0&quot; encoding=&quot;UTF-8&quot; standalone=&quot;no&quot;?&gt;&#10;&lt;runtimeClasspathEntry internalArchive=&quot;/tuxguitar-gm-settings/target/tuxguitar-gm-settings-SNAPSHOT.jar&quot; path=&quot;5&quot; type=&quot;2&quot;/&gt;&#10;"/>
+        <listEntry value="&lt;?xml version=&quot;1.0&quot; encoding=&quot;UTF-8&quot; standalone=&quot;no&quot;?&gt;&#10;&lt;runtimeClasspathEntry internalArchive=&quot;/tuxguitar-gm-utils/target/tuxguitar-gm-utils-SNAPSHOT.jar&quot; path=&quot;5&quot; type=&quot;2&quot;/&gt;&#10;"/>
```
guiv42, line 21: see my comments in #276: it does not solve the issue (at least in my config)

guiv42:
> In the run configuration->Dependencies select the Classpath Entries, click Add External Jars....
Now go to tuxguitar/desktop/build-scripts/tuxguitar-linux-swt-x86_64/target/tuxguitar-SNAPSHOT-linux-swt-x86_64/lib, and add all of them. This works at on my PC

It does not work "as is" in my configuration, but I think you've found the correct path to a solution. The build directives in the pom file copy the jar files of the different plugins to a dedicated folder : see `<outputDirectory>` tag defined in [pom file](https://github.com/helge17/tuxguitar/blob/08d5e6c7969a3ffc6b8ed9eb1f755508059fd4c4/desktop/build-scripts/tuxguitar-linux-swt-x86_64/pom.xml#L160). So, in `Run configuration -> Dependencies`, select `Classpath Entries`, click `Add External jars...`, browse to `tuxguitar/desktop/build-scripts/tuxguitar-linux-swt-x86_64/target/tuxguitar-SNAPSHOT-linux-swt-x86_64/share/plugins/` and add all (?) of them.
With this I get an error when debugging starts because in my configuration one plugin doesn't work properly (`java sound api cannot be loaded`). All the other plugins work properly. Removing `tuxguitar-jsa` solves the error.

I don't know what differs between our configurations, I'm working on Linux. I can understand why adding jar files from **share/plugins** can solve the issue, but I really don't understand why files from **lib** folder should change something.

profOnno:
Hmm interesting, 

I only seem to need `gervil` from the lib folder, else it will start but the console is full of errors.
If i want to use musicxml export i need to have `tuxguitar-gm-utils.jar` from the lib folder. It is an import in MusicXML writer. I also import tuxguitar-lib files... like TGTrack.java but i don't have to add this to the dependencies.

I will try to make a new work space, and do a fresh install, trying to follow the install instructions exactly.

guiv42:
I'm starting  to understand. In fact we had already identified the need to add some dependencies, see #175 (quite a long discussion, now closed): at least swt and ui-toolkit-swt were required, or else the ui did not even start. So we added these 2, delivered a debug configuration and stopped there. These 2 jars are copied to `lib` folder by the directives of the pom file. But we did not think that many other jars could be missing in the debug config (at that period we were already happy to have succeeded launching the debugger!).
So, in debug dependencies we should probably replace the 2 jars present in delivered config by all jars from both `lib` and `share\plugins`

guiv42:
@profOnno, could you please test the following procedure?
initial conditions:
- clean, empty Eclipse configuration: nothing in Eclipse workspace
- no folder ~/.config/tuxguitar
- tuxguitar already built from a console: folder `desktop/build-scripts/tuxguitar-linux-swt-x86_64/target/tuxguitar-9.99-SNAPSHOT-linux-swt-x86_64/` exists and contains an executable app.

Procedure in Eclipse:
1. File/Import/Maven/Existing Maven project, then browse to `desktop/build-scripts/tuxguitar-linux-swt-x86_64`
2. Run/Run configurations/select "Java Application" and right-click "new configuration"
3. in "Main" tab, project: browse to `tuxguitar`, and Main Class: browse to `org.herac.tuxguitar.app.TGMainSingleton`
4. In Arguments tab, "VM arguments" field, type: `-Dtuxguitar.share.path="(absolute path to your tuxguitar folder)/desktop/build-scripts/tuxguitar-linux-swt-x86_64/target/tuxguitar-9.99-SNAPSHOT-linux-swt-x86_64/share/"`
5. In Dependencies tab, delete Classpath Entries: `tuxguitar` and `Maven dependencies`
6. In Dependencies tab, Classpath Entries, Add External JARs..., browse to `desktop/build-scripts/tuxguitar-linux-swt-x86_64/target/tuxguitar-9.99-SNAPSHOT-linux-swt-x86_64/lib` and add everything

Then "Apply", and "Run"
With this, it seems to work rather correctly in my configuration. I still get an unexplained exception when closing app, but this should be acceptable for most debugging use cases.

My understanding:
- normal app: build directives in `tuxguitar-linux-swt-x86_64` pom file copy the required dependencies (jars, translations, icons, etc) at specific locations. These locations are provided to TuxGuitar via parameters of the command line (see `tuxguitar.sh` launch script): some are provided to the java environment (classpath), some are provided directly to TuxGuitar (e.g. "/share" folder, see step 4)
- Eclipse tries to recreate classpath from pom file of `tuxguitar` project (where the "main" is located), so it's missing all dependencies of `tuxguitar-linux-swt-x86_64`. You can see this just before step 5, in the window to create run configuration, there is a button "Show Command Line", it shows clearly the classpath parameter.
I could not find a way to configure Eclipse exactly in the same manner, so the procedure above remains an approximation.

PR #282, opened 2024-03-14
Adds basic musicxml lyric support

profOnno:
Adds basic musicxml lyric support

my notes on the state machine:
[Quick sheets.pdf](https://github.com/helge17/tuxguitar/files/14605513/Quick.sheets.pdf)

Not sure about the naming of my methods and classes.

Now it will export the lyrics twice, to the main track and to the tablature. Not sure about this, maybe only on the main track is better.

In the `getLyricSlice` function I'm using `Arrays.stream`,  the are other ways to do this if this is only supported in newer Java versions.

I did add javadoc compatible comments. I let it generate in `doc/java_docs` but the config for this is not in this request.

Take your time with the review. Java is a language I didn't touch in a decade, so there are probably better ways to do some things.

Cheers,

Menno

### DIFF
--- a/desktop/TuxGuitar-musicxml/src/org/herac/tuxguitar/io/musicxml/MusicXMLLyricWriter.java
+++ b/desktop/TuxGuitar-musicxml/src/org/herac/tuxguitar/io/musicxml/MusicXMLLyricWriter.java
@@ -1,0 +1,193 @@
+package org.herac.tuxguitar.io.musicxml;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import org.herac.tuxguitar.graphics.control.TGMeasureImpl;
+import org.herac.tuxguitar.song.models.TGTrack;
+
+/**
+ * MusicXMLLyricWriter is used to convert TGLyric to multiple MusicXMLMeasureLyric chunks.
+ * @author Menno Vossen
+ */
+public class MusicXMLLyricWriter {
+	/**
+	 * Syllabic enum.
+	 * See the MusicXML specification: <a href="https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/syllabic/">syllabic</a>
+	 */
+	public enum Syllabic {
+		NONE,
+		SINGLE,
+		BEGIN,
+		MIDDLE,
+		END;
+		
+		@Override
+		public String toString() {
+			switch(this.ordinal()) {
+			case 0:
+				return "none"; // NOTE: not a official syllabic but used in state machine.
+			case 1:
+				return "single";
+			case 2:
+				return "begin";
+			case 3:
+				return "middle";
+			case 4:
+				return "end";
+			default:
+				return "single";
+			}
+		}
+	}
+	/**
+	 * MusicXMLMeasureLyric contains the syllabic and text for one note.
+	 * See the MusicXML specification:
+	 * <a href="https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/lyric/">lyric</a>
+	 */
+	public class MusicXMLMeasureLyric {
+		public Syllabic syllabic;
+		public String text;
+		
+		/**
+		 * Class constructor.
+		 * @param syllabic
+		 * @param text
+		 */
+		public MusicXMLMeasureLyric(Syllabic syllabic, String text) {
+			this.syllabic = syllabic;
+			this.text = text;
+		}
+	}
+	/**
+	 * Helper function.
+	 * @param string
+	 * @return
+	 */
+	private static String removeLastChar(String string) {
+	    return (string == null || string.length() == 0)
+	    		? null : (string.substring(0, string.length() - 1));
+	}
+	
+	/**
+	 * Helper function.
+	 * The lyrics in TGLyric are for one track. The lyricStrings are lyric per note.
+	 * @param lyricStrings		string array source
+	 * @param start				start index for the slice
+	 * @param number			number of strings to get
+	 *
+	 * @return 	The string Array is used for one measure.
+	 */
+	private static String[] getLyricSlice(String[] lyricStrings, int start, int number) {
+		try {
+			return Arrays.stream(lyricStrings, start, start+number).toArray(String[]::new);
+		} catch (ArrayIndexOutOfBoundsException exception) {
+			// ignore
+			// more lyrics than notes!?
+		}
+		
+		// we need to return a default
+		String[] empty = {""};
+		return empty;
+	}
+	
+	/**
+	 * Holds the lyrics for the complete track.
+	 */
+	private String[] lyrics;
+	/**
+	 * Used to keep track off our position in the lyrics array.
+	 */
+	private int lyricIndex;
+	/**
+	 * Used to store the start of lyrics offset in Measures.
+	 */
+	private int lyricFrom;
+	/**
+	 * Used in the generateLyricList method, which is a state machine.
+	 */
+	private Syllabic syllabicState;
+	
+	
+	/**
+	 * MusicXMLLyricWriter Constructor.
+	 * @param track 	Holds track currently being processed.
+	 */
+	public MusicXMLLyricWriter(TGTrack track) {
+		lyricFrom = track.getLyrics().getFrom();
+		lyrics = track.getLyrics().getLyricBeats();
+		lyricIndex = 0; // in lyrics (Strings)
+		syllabicState = Syllabic.NONE;
+	}
+	/**
+	 * State machine, use only sequential calls to get lyrics from track.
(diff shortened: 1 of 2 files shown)


guiv42:
Thanks for the contribution!

> Not sure about the naming of my methods and classes.

No problem. Nobody can be sure of anything, since there's no such thing as coding rules or naming conventions. We'll need to do this one day.

> Now it will export the lyrics twice, to the main track and to the tablature. Not sure about this, maybe only on the main track is better.

I've got to test this, but intuitively I would think only the main track would be more readable.

> Take your time with the review. Java is a language I didn't touch in a decade, so there are probably better ways to do some things.

I'm definitely no java expert. So the added value of my review should be limited to the interactions of your code with other parts/modules of TuxGuitar. I can also do some testing, but not much more.

desktop/TuxGuitar-musicxml/src/org/herac/tuxguitar/io/musicxml/MusicXMLLyricWriter.java
```diff
@@ -0,0 +1,208 @@
...
+	
+	/**
+	 * Helper function.
+	 * @param string
+	 * @return
+	 */
+	private static String removeLastChar(String string) {
+	    return (string == null || string.length() == 0)
+	      ? null 
+	      : (string.substring(0, string.length() - 1));
```
guiv42, line 73: minor remark: this could be re-indented (replace leading spaces with tabs)

desktop/TuxGuitar-musicxml/src/org/herac/tuxguitar/io/musicxml/MusicXMLWriter.java
```diff
@@ -154,12 +156,16 @@ private void writeTrack(TGTrack track, Node parent, boolean isTablature){
 		Node part = this.addAttribute(this.addNode(parent,"part"), "id", "P" + track.getNumber() + (isTablature ? TABLATURE_SUFFIX : ""));
 		
 		TGMeasure previous = null;
+	
+		MusicXMLLyricWriter lyricWriter = new MusicXMLLyricWriter(track);
+		int lyricFrom = track.getLyrics().getFrom();
```
guiv42, line 161: unused variable, can be deleted

desktop/TuxGuitar-musicxml/src/org/herac/tuxguitar/io/musicxml/MusicXMLWriter.java
```diff
@@ -339,10 +358,31 @@ private void writeBeats(Node parent, TGMeasure measure, boolean isTablature){
 						this.addNode(technicalNode,"string", Integer.toString( note.getString() ));
 					}
 				}
+
+				try {				
+					MusicXMLMeasureLyric measureLyric = lyrics[lyricIndex++];
+
+					// NOTE: Now its in both tablature and normal				
+					writeLyric(noteNode, measureLyric);
```
guiv42, line 366: as suggested in your PR, I think exporting lyrics only under score (and not under tablature) makes the result more readable. Just adding a condition `if (!isTablature)` should do the job.

desktop/TuxGuitar-musicxml/src/org/herac/tuxguitar/io/musicxml/MusicXMLLyricWriter.java
```diff
@@ -0,0 +1,208 @@
...
+	 */
+	private Syllabic syllabicState;
+	
+	
+	/**
+	 * MusicXMLLyricWriter Constructor.
+	 * @param track 	Holds track currently being processed.
+	 */
+	public MusicXMLLyricWriter(TGTrack track) {
+		//track = _track; // do we have to specify this?
```
guiv42, line 124: is this question still open? It seems it isn't, so I think this line can be deleted

desktop/TuxGuitar-musicxml/src/org/herac/tuxguitar/io/musicxml/MusicXMLLyricWriter.java
```diff
@@ -0,0 +1,208 @@
...
+		
+		
+		String[] measureLyricStrings = getLyricSlice(lyrics, lyricIndex, number);				
+		this.lyricIndex += number;
+											
+		Syllabic syllabic = Syllabic.NONE;
+			
+		
+		// TODO: handle notes without lyrics here?
+		// or keep the lyric "", and filter out in write beats
```
guiv42, line 154: Actual result seems pretty good, I would not change it.

desktop/TuxGuitar-musicxml/src/org/herac/tuxguitar/io/musicxml/MusicXMLLyricWriter.java
```diff
@@ -0,0 +1,208 @@
...
+					syllabic = Syllabic.NONE;
+					syllabicState = Syllabic.NONE;
+					break;
+			}
+
+			measureLyrics.add(new MusicXMLMeasureLyric(syllabic, measureLyricStrings[i]));			
+		}		
+		
+		MusicXMLMeasureLyric[] measureLyricArray = new MusicXMLMeasureLyric[measureLyrics.size()];
+		//measureLyricArray = (MusicXMLMeasureLyric[]) measureLyrics.toArray();
```
guiv42, line 204: this line can be deleted

guiv42 (commented):
I find the result is really good, great job, thanks!

As you see, my remarks are essentially formal, I only suggest minor modifications.
Exporting lyrics only with score (and not with tablature) seems to me the only significant evolution.
And could you please re-indent the 2 files? (e.g. several lines with useless tabs at end of line)

(by the way, when testing your modification I detected a bug already present before: tab shall not be exported for drums track. I will fix this after this PR is merged)

profOnno:
Thank you for the review. I agree with your remarks, and will adjust the code accordingly.

desktop/TuxGuitar-musicxml/src/org/herac/tuxguitar/io/musicxml/MusicXMLWriter.java
profOnno: Changing this (wait for the pull). There was a bug when using chords. The new code will show this. So the lyric will be attached to the first note on a chord, or just on the note if there is no chord.

profOnno:
Pfff... Trying to understand the git workflow to rebase is hard work :sweat_smile:.
But I think it worked.

desktop/TuxGuitar-musicxml/src/org/herac/tuxguitar/io/musicxml/MusicXMLWriter.java
```diff
@@ -307,27 +315,44 @@ private void writeDirection(Node parent, TGMeasure measure, TGMeasure previous){
...
 				int noteCount = voice.countNotes();
+				
 				for(int n = 0; n < noteCount; n ++){
 					TGNote note = voice.getNote( n );
 					
 					Node noteNode = this.addNode(parent,"note");
+					
 					int value = (beat.getMeasure().getTrack().getString(note.getString()).getValue() + note.getValue());
 					
 					if(n > 0){
```
guiv42, line 344: replace lines 344 to 360 by this code:

```
					if(n > 0){
						this.addNode(noteNode,"chord");
					}
					
					Node pitchNode = this.addNode(noteNode,"pitch");
					this.writeNote(pitchNode, "", value, ks);
					this.writeDuration(noteNode, voice.getDuration(), note.isTiedNote());
					
					if ((n==0) && !isTablature) {
						// Attach lyric to the first note
						try {
							MusicXMLMeasureLyric measureLyric = lyrics[lyricIndex++];
							writeLyric(noteNode, measureLyric);
						} catch (Exception e) {
							// ignore
							// can be out of bound? when there are more lyrics than text
							// can be null if there is an offset
						}
					}
```

guiv42 (changes requested):
See my comment, it fixes issue.

profOnno:
Did it a little bit different but it passed the xmllint. Now without rebase. Thanks for the explanation on how to do this.
I'll try to add xmllint in the test maven test, in a new pull request (if I get it working)

guiv42:
> Did it a little bit different but it passed the xmllint.

It's ok now, thanks.

> Do you have a blueprint of how to add a test? I added a test folder in next to the src and used eclipse to add a junit test.

No, but I would really like to, that's yet another thing I have to learn. I introduced the first tests in the code base recently, without a clear knowledge of the correct way to do it, so it's definitely not clean. If you can find how to do that, I will refactor org.herac.tuxguitar.test package.

Issue #283, opened 2024-03-14
can't open .gp files any more


guiv42:
TuxGuitar desktop, SWT, Linux

popup window: `org/apache/commons/io/input/NullInputStream`

Looks like a dependency issue, problem appears with 2f681c7d5c82509fbf6d1ef573ed505070e905ab

edit: *very* annoying... my distro also updated commons-compress to 1.26, so TuxGuitar 1.6.1 installed by the distro's repo does not open .gp files any more.

```
org.herac.tuxguitar.action.TGActionException: org/apache/commons/io/input/NullInputStream
        at org.herac.tuxguitar.app.action.impl.file.TGReadURLAction.processAction(TGReadURLAction.java:35)
        at org.herac.tuxguitar.editor.action.TGActionBase.execute(TGActionBase.java:24)
        at org.herac.tuxguitar.action.TGActionManager.execute(TGActionManager.java:67)
        at org.herac.tuxguitar.editor.action.TGActionProcessor.processOnCurrentThread(TGActionProcessor.java:75)
        at org.herac.tuxguitar.editor.action.TGActionProcessor.processOnCurrentThread(TGActionProcessor.java:88)
        at org.herac.tuxguitar.editor.action.TGActionProcessor$1.run(TGActionProcessor.java:98)
        at org.herac.tuxguitar.thread.TGPooledThreadHandler.processNext(TGPooledThreadHandler.java:66)
        at org.herac.tuxguitar.thread.TGPooledThreadHandler.run(TGPooledThreadHandler.java:86)
        at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: org.herac.tuxguitar.action.TGActionException: org/apache/commons/io/input/NullInputStream
        at org.herac.tuxguitar.editor.action.file.TGReadSongAction.processAction(TGReadSongAction.java:48)
        at org.herac.tuxguitar.editor.action.TGActionBase.execute(TGActionBase.java:24)
        at org.herac.tuxguitar.action.TGActionManager.execute(TGActionManager.java:67)
        at org.herac.tuxguitar.app.action.impl.file.TGReadURLAction.processAction(TGReadURLAction.java:33)
        ... 8 more
Caused by: org.herac.tuxguitar.io.base.TGFileFormatException: org/apache/commons/io/input/NullInputStream
        at org.herac.tuxguitar.io.gpx.v7.GPXInputStream.read(GPXInputStream.java:28)
[...]

PR #284, opened 2024-03-14
fix dependency

guiv42:
see #283

following 2f681c7d5c82509fbf6d1ef573ed505070e905ab
commons-compress update now requires explicit dependency to commons-io

needed to open .gp files

tested on Linux (openSUSE Tumbleweed) and Windows 10, SWT and JFX
tested on FreeBSD, SWT only
built but not tested on FreeBSD, JFX

not built not tested on macOS

@helge17 : since it modifies the pom files, I would appreciate if you could test in macOS configuration (I prefer not to merge this PR myself)


helge17:
Thanks! Also works on MacOS.

PR #285, opened 2024-03-15
Added documentation for MusicXML export

profOnno:
Added documentation for MusicXML export. I did edit the html file. If it's generated from an other source, I need to revisit it.


guiv42:
to be merged after #282

PR #291, opened 2024-03-18
Add recurse mkdir

G-eos:
mkdir fails on a fresh cloned repository in my case. Adding -p to mkdir fix it.

### DIFF
--- a/misc/build_tuxguitar_from_source.sh
+++ b/misc/build_tuxguitar_from_source.sh
@@ -467,7 +467,7 @@ for GUI_TK in swt jfx; do
   mvn --batch-mode -e clean verify
 
   # Copy locally installed openjdk (from Homebrew) to get it integrated in the APP.TAR.GZ packages
-  mkdir target/tuxguitar-$TGVERSION-macosx-$GUI_TK-cocoa-$BUILD_ARCH.app/Contents/MacOS/jre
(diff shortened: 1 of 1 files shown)


helge17:
I assume you ran the script with something like ./misc/build_tuxguitar_from_source.sh -m on MacOS locally, right?

The build script is very specific to my build environment: currently the source code is manipulated with sed on Linux, then transferred to MacOS via rsync and built there with the -m option.

There is deliberately no -p option after mkdir: If the parent directory does not exist, then something has gone wrong. That's the reason why I'm not willing to merge this PR.

If you add the -p option, it still doesn't work locally on MacOS: the build is successful, but you end up with a broken app package.

But I agree that it would be nice if the -m (MacOS) and -b (BSD) options would also work locally on the target OS. I opened issue #298 for that.

G-eos:
Thanks to have taken time to answer clearly. I understand this is a good way to detect that something goes wrong.
Yes it is with -m
For information it builds two files into:
tuxguitar/desktop/build-scripts/tuxguitar-macosx-swt-cocoa-x86_64/target

One has jre the other no jre. Mixing the two makes a tuxguitar which seems to work. No intensive test.

I understand that this is not the clean way to build. I will follow the Contribute.md file to have a clean way to build and test in real condition. Thanks again to take time to have a look and answer.

PR #304, opened 2024-03-22
Configurable tuning presets in xml file using priority attributes

tdan:
- Add an optional attribute priority to tuning presets in tuning.xml file
- Add an integer attribute priority in TGTuning class, with value imported from xml file content
- Implement Comparable<TGTuning> interface in TGTuning class, so instances can be sorted by priority order
- Add method getPrioritizedTgTunings(), which returns all TG tuning presets with a defined priority attribute, discard tuning presets with no priority.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ desktop/build-scripts/native-modules/tuxguitar-synth-vst-*/include/
 .project
 android/build-scripts/tuxguitar-android/local.properties
 android/build-scripts/tuxguitar-android/.idea/
+desktop/build-scripts/*
+00-Binary_Packages/
--- a/common/TuxGuitar-lib/src/org/herac/tuxguitar/song/helpers/tuning/TuningManager.java
+++ b/common/TuxGuitar-lib/src/org/herac/tuxguitar/song/helpers/tuning/TuningManager.java
@@ -13,72 +13,92 @@ import org.herac.tuxguitar.util.singleton.TGSingletonUtil;
 import java.io.File;
 import java.io.FileInputStream;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 public class TuningManager {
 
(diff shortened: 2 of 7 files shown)


guiv42 (changes requested):
Okay, there are quite a few blocking items here

1. The most critical one: only modify the minimum set of files required to implement the expected evolution. Especially do **not** modify `.gitignore`. In this case, I just cannot run the app from your branch (it builds but fails to start). Also it hides probably some modifications that you have done in other parts of the code base.
Basically for this evolution, modifying *only* these files should do the job: TuningManager.java, TuningReader.java, TGTuning.java, MidiSongReader.java, tunings.xml. No less, no more.
I don't know why the scope of your modifications was wider. Don't hesitate to ask for support if needed.
2. don't push from your master branch. In case your pull request is not merged (like this one), you will need to re-synchronize your master branch. And please squash your commits before submitting a PR (the intermediate steps of your development should not appear in main branch history). One classical way to do this:
- from master branch, create 2 branches : "feature" (here 269) and "dev"
- work in your "dev" branch, with as many commits as you like
- when you're ready, `merge squash` "dev" into "feature", and push "feature" branch to create PR. So in your PR, only one commit appears.

3. Please don't break indentation, this makes the code unreadable

4. About the functional evolution: most of the job is correctly done, except the last part (midi import) where the added value is expected. Objective of this issue is to remove redundancy between java code and xml, yet it is still present. See my comments in MidiFileReader, this should be easy to fix I think. All the other modified files are OK :)

So, this PR cannot be merged. But there's not much job to do to make this acceptable. Here's the best way to go forward I think:
- do not try to update this PR, there are some modifications in the commit history that I don't want to appear in master's history
- create a new branch (or 2 as suggested above) from the last version of master
- copy into this branch the few modified files listed above
- apply in MidiFileReader the required correction
- create a new PR

tdan:
Hey @guiv42 I really appreciate your feedback! This is my first PR ever so I have a lot of things to learn :).

Gonna work on it ASAP.

Issue #312, opened 2024-03-29
Windows Defender Malware?


GruNostalgia:
Windows 11 Pro 23H2 Here
Installed the windows version: tuxguitar-1.6.1-windows-swt-x86_64-installer.exe 
Defender says it picked up malware for - tuxguitar-synth-vst.exe
VirusTotal also gives malware detection
![tuxguitar](https://github.com/helge17/tuxguitar/assets/80476754/5753c072-195c-40d3-a305-18a5a92bcbd1)

I'm not new to false positives but this seems extreme. Any clarification would be appreciated, thanks!

guiv42:
This one is unexpected, as far as I know this module has been there for a while. Thanks for raising the issue.
It really looks like a false positive.

Now, considering this plugin:
- vst2 has been deprecated for years
- licensing is not so clear (to me at least)
- it complexifies build procedure (e.g. see #310)
- its build procedure is a *hack*, and explicitly [documented](https://github.com/helge17/tuxguitar/blob/ca8bec5cfd865474a3dcc09869c9bc148f8b4d91/INSTALL.md?plain=1#L55) as such
- it's already been removed from tuxguitar packages delivered by Linux distros since a long time
- my understanding is that tuxguitar-synth-lv2 provides similar features (not 100% sure though)
- last but not least: I'm not fond of archeology

So, here's my proposal: keep it simple and be efficient, delete tuxguitar-synth-vst

If users complain I would consider to build/deliver it independently, from another repo.

@helge17: what's your opinion?

helge17:
In fact, Windows Defender classifies this file as "Trojan:Win32/Wacatac.A!ml" and deletes it.

Only the tuxguitar-synth-vst.exe from version 1.6.1 is classified as malware, not the one from version 1.5.6, 1.6.0 or yesterday's build 2024-03-28-master. I have checked the sources of all versions: They are all identical. "My" versions 1.6.0, 1.6.1, 2024-03-28-master were all built on the same Linux system, so I also think this is a false positive.

I have just sent the file to Microsoft for rescanning. As soon as I hear back, I will let you know.

I also think we should soon remove tuxguitar-synth-vst, but that will only "help" future versions. I would prefer to resolve this issue with version 1.6.1, otherwise I am considering removing this version from the releases.

guiv42:
> I would prefer to resolve this issue with version 1.6.1, otherwise I am considering removing this version from the releases

OK, clear.
[...]

PR #315, opened 2024-03-31
feat: add functionality to save and load preset ranges in play mode

iamn1ck:
Hi, I use tuxguitar regularly and the play mode range selector is not always the best user experience. I added a new UI state to save and load preset ranges for the tab. The feature is almost implemented correctly, except after confirming your selection in the play mode dialog the saved preset are forgotten (see the video attached). It would also be very helpful if it could persist the presets when the app is closed and opened again, so the user does not have to keep creating the same presets every time. Any pointers on how to implement this would be appreciated.

https://github.com/helge17/tuxguitar/assets/3345597/6e2a4446-746c-4a31-9e5a-00438919efc1


common/TuxGuitar-lib/src/org/herac/tuxguitar/player/base/MidiPlayerLoop.java
```diff
@@ -0,0 +1,64 @@
...
+import org.herac.tuxguitar.song.models.TGSong;
+import org.herac.tuxguitar.song.models.TGString;
+import org.herac.tuxguitar.song.models.TGTrack;
+import org.herac.tuxguitar.thread.TGThreadLoop;
+import org.herac.tuxguitar.thread.TGThreadManager;
+import org.herac.tuxguitar.util.TGContext;
+import org.herac.tuxguitar.util.TGLock;
+import org.herac.tuxguitar.util.TGMessagesManager;
+import org.herac.tuxguitar.util.singleton.TGSingletonFactory;
+import org.herac.tuxguitar.util.singleton.TGSingletonUtil;
```
guiv42, line 26: remove all unused `import` statements

desktop/TuxGuitar/src/org/herac/tuxguitar/app/view/dialog/transport/TGTransportModeDialog.java
```diff
@@ -205,6 +217,66 @@ public void show(){
 		this.loopEHeader = uiFactory.createDropDownSelect(rangeGroup);
 		rangeLayout.set(this.loopEHeader, 2, 2, UITableLayout.ALIGN_FILL, UITableLayout.ALIGN_FILL, true, true);
 		mHeaderRangeStatus.addControl(this.loopEHeader);
+
+
+		UILabel presetLabel = uiFactory.createLabel(rangeGroup);
+		presetLabel.setText(TuxGuitar.getProperty("transport.mode.preset") + ":");
```
guiv42, line 223: `transport.mode.presets` with a trailing `s` in translation file

desktop/TuxGuitar/src/org/herac/tuxguitar/app/view/dialog/transport/TGTransportModeDialog.java
```diff
@@ -380,6 +467,27 @@ public void onSelect(UISelectionEvent event) {
...
+				this.presetList[i] = new MidiPlayerLoop();
+			}
+
+			this.presetSelect.removeItems();
+
+			for (int i=0; i < presetList.length; i++) {
+				this.presetSelect.addItem(new UISelectItem<MidiPlayerLoop>(presetList[i].getTitle()));
+			}
+		}
+	}
```
guiv42, line 490: dead code

desktop/TuxGuitar/src/org/herac/tuxguitar/app/view/dialog/transport/TGTransportModeDialog.java
```diff
@@ -63,6 +68,13 @@ public void show(){
 		final UITableLayout dialogLayout = new UITableLayout();
 		final UIWindow dialog = uiFactory.createWindow(uiParent, true, false);
 		final TGBeatRange beats = this.context.getAttribute(TGDocumentContextAttributes.ATTRIBUTE_BEAT_RANGE);
+		// final MidiPlayerLoop[] loops = new MidiPlayerLoop[5];
+
+
+		// for (int i=0; i < loops.length; i++) {
+		// 	loops[i] = new MidiPlayerLoop();
+		// }
+
```
guiv42, line 77: delete commented code

desktop/TuxGuitar/src/org/herac/tuxguitar/app/view/dialog/transport/TGTransportModeDialog.java
```diff
@@ -205,6 +217,66 @@ public void show(){
...
+		rangeLayout.set(this.presetSelect, 3, 2, UITableLayout.ALIGN_FILL, UITableLayout.ALIGN_FILL, true, true);
+		mHeaderRangeStatus.addControl(this.presetSelect);
+		// this.presetSelect.addSelectionListener(new UISelectionListener() {
+		// 	public void onSelect(UISelectionEvent event) {
+		// 		// load beginning and end
+		// 		System.out.print("\n\nselected" + mode.getLoopSHeader() + " " + mode.getLoopEHeader());
+				
+		// 		System.out.print("\n\nselected" + presetSelect.getSelectedValue().getTitle());
+		// 	}
+		// });
```
guiv42, line 250: delete commented code

desktop/TuxGuitar/src/org/herac/tuxguitar/app/view/dialog/transport/TGTransportModeDialog.java
```diff
@@ -205,6 +217,66 @@ public void show(){
...
+
+
+		UILabel presetLabel = uiFactory.createLabel(rangeGroup);
+		presetLabel.setText(TuxGuitar.getProperty("transport.mode.preset") + ":");
+		rangeLayout.set(presetLabel, 3, 1, UITableLayout.ALIGN_FILL, UITableLayout.ALIGN_CENTER, false, true);
+		mHeaderRangeStatus.addControl(presetLabel);
+		
+
+		UILabel presetNameLabel = uiFactory.createLabel(rangeGroup);
+		presetNameLabel.setText("name label" + ":");
```
guiv42, line 229: use text property from translation file (not hard-coded)

desktop/TuxGuitar/src/org/herac/tuxguitar/app/view/dialog/transport/TGTransportModeDialog.java
```diff
@@ -205,6 +217,66 @@ public void show(){
...
+		buttonLoad.setText(TuxGuitar.getProperty("transport.mode.load"));
+		buttonLoad.setDefaultButton();
+
+
+		UIButton buttonSave = uiFactory.createButton(rangeGroup);
+		buttonSave.setText(TuxGuitar.getProperty("transport.mode.save"));
+		buttonSave.setDefaultButton();
+		buttonSave.addSelectionListener(new UISelectionListener() {
+			public void onSelect(UISelectionEvent event) {
+				System.out.print("\n\nsaving beginning and endsss" + presetNameText.getText() + "" +  loopSHeader.getSelectedValue() + " " + loopSHeader.getSelectedValue());
```
guiv42, line 265: remove all debug logs (`System.out.print`)

guiv42 (changes requested):
I agree about the play mode range selector, and I like the idea. Thanks for the proposal.

There are a few functional issues in your implementation:
1. Buttons
I don't understand the logic of buttons: what is "load preset" supposed to do? I would think that the drop-down list would already be populated when opening dialog (see below), then selecting a preset in drop-down list would "load" it.
"Save Preset" button should be enabled only if "name label" is not empty.
It's only possible to add presets, not to remove them. A "Clean" or "Delete" button is missing.

2. A crash scenario:
- open new song, type F9 to open dialog
- click "play looped"
- click "Load Preset" -> crash (null pointer exception)


Then there's another problem: what if the song is edited?
Example use case: a loop preset is defined on interval [4,6], and user edits song
- insert new measure before 4 -> loop preset should now correspond to [5,7]
- delete measure 5 -> [6,7]
- insert 1 measure between 6 and 7 -> [6,8]
- delete measures 6 to 8 -> loop preset should be deleted
- etc.

Fundamentally, a "loop preset" corresponds to an interval of measures, and this should be a property of song object. I think it should store *references* to measures (more precisely to measure headers, common to all tracks) and not measure *numbers*, so that song edition can be considered.

A possible way forward:
- replace `MidiPlayerLoop` class by a class `TGMeasuresInterval`, in package org.herac.tuxguitar.song.models, with similar attributes (title, start, end)
- in this class, replace measure *numbers* (start,end) by *references* to TGMeasureHeader objects. It's easy to get a header from a measure number, and vice-versa.
- store a list of `TGMeasuresInterval` objects in TGSong. A `List<>` would be better than an array with hard-coded size (visible in your commented code).
- in dialog, retrieve this list from song to initialize the different fields. A bit tricky, since some references may be null, or not null but no more linked to current song (see the song edition scenarios above)

Important note: to add remanence, it is required to store this list of intervals in song file. So it requires an evolution of the .tg file format, breaking compatibility. Old versions of TuxGuitar would not be able to open files encoding such data.
So I suggest to target `tuxguitar-next` branch instead of `master`.
@iamn1ck: this branch targets the first version of TuxGuitar breaking file fomat compatibility. It's not planned yet, but it needs to be done some day (also required by other feature requests).

iamn1ck:
Thanks for the feedback and suggestions. I went ahead and implemented most of the changes already. 

I have renamed the model as you suggested, to TGMeasuresInterval, and updated the UI to be more intuitive. I was also able to persist the list of TGMeasuresInterval on the TGSong so you can open and close the dialog and keep the presets. 

The only thing missing is persisting the presets on the tgsong file, but I'm not sure where to add that.

guiv42:
> Thanks for the feedback and suggestions. I went ahead and implemented most of the changes already.

OK, good. It might take some time for me to review these changes, many open topics in the todo list currently.

> The only thing missing is persisting the presets on the tgsong file, but I'm not sure where to add that.

This is clearly an open point, and it's too early to answer your question.
Currently our plan is the following:
1. finalize delivery of version 1.6.2 (this release was unexpected, see #312)
2. still some feature requests / bug fixes to implement in master branch, targeting a version 1.6.3 without any modification of file format
3. implement other feature requests requiring file format evolution. Possibly for a version 2.0? Still in discussion

I cannot give you any date for these steps. This is a spare time project, so globally we're progressing slowly (but we're progressing!). And this plan may change. Refer to #41 to follow the discussions on dev next steps.
Between steps 2 and 3, we will need to open the discussion about the new file format. Most probably we will not keep the existing custom binary (and undocumented) format.

About your new feature: is there still some added value if the loop intervals are not persistent? I don't really know.
If persistence is required, then we should target tuxguitar-next branch, and wait some time (step 3 above) to merge into master.

iamn1ck:
I think this is a good starting point, but unless the preset intervals can be saved on the tgsong file (or elsewhere) then the feature is not too helpful. I will start to rebase my changes to the tuxguitar-next branch and then we can discuss how to approach the file format changes

guiv42:
Just a suggestion, proposed by another user on a french-speaking guitar [forum](https://www.forum-guitare.fr/viewtopic.php?t=58851&start=56), completely independently from your work: it might even be more interesting to store in TGSong object *all settings* from player mode dialog.

Use case : working simultaneously with 2 songs.
E.g. I'm working on song 1, I set tempo 90%, I select song2 to look at something, then come back to song1 -> currently tempo is reset to 100%

So the best solution could be to store all player mode settings in song object (including the current loop preset, not only the presets list), but only to persist some settings in the song.tg file (e.g. loop presets)

What do you think?

desktop/TuxGuitar/src/org/herac/tuxguitar/app/view/dialog/transport/TGTransportModeDialog.java
```diff
@@ -277,6 +310,61 @@ public void changeTransportMode() {
...
+			public void onSelect(UISelectionEvent event) {
+				TGTransportModeDialog.this.onRemovePreset();
+			}
+		});
+		
+		UIButton buttonEdit = factory.createButton(buttonsPanel);
+		buttonEdit.setImage(TGIconManager.getInstance(this.context.getContext()).getListEdit());
+		buttonEdit.addSelectionListener(new UISelectionListener() {
+			public void onSelect(UISelectionEvent event) {
+				TGTransportModeDialog.this.onLoadPreset();
```
guiv42, line 357: I don't understand the behavior of "edit" button (calling "onLoad"?)
I assume it's still work in progress

desktop/TuxGuitar/src/org/herac/tuxguitar/app/view/dialog/transport/TGTransportModeDialog.java
```diff
@@ -440,4 +530,83 @@ public void onSelect(UISelectionEvent event) {
...
+			this.presetTable.addItem(item);
+		}
+		
+		if( selection != null ) {
+			this.presetTable.setSelectedValue(selection);
+		}
+	}
+	
+
+	private void updateTuningControls() {
```
guiv42, line 563: why "tuning"? (copy-pasted from tuning dialog?)
-> rename loopPreset

This method should be deleted (why a method just to call another one?)

desktop/TuxGuitar/src/org/herac/tuxguitar/app/view/dialog/transport/TGTransportModeDialog.java
```diff
@@ -440,4 +530,83 @@ public void onSelect(UISelectionEvent event) {
...
+		TGMeasuresInterval preset = this.presetTable.getSelectedValue();
+		if( preset != null ) {
+			removePreset(preset);
+		}
+	}
+
+
+	private void onAddPreset() {
+		TGSong song = TGTransportModeDialog.this.context.getAttribute(TGDocumentContextAttributes.ATTRIBUTE_SONG);
+		TGMeasureHeader startHeader = song.getMeasureHeader( this.loopSHeader.getSelectedValue() - 1 );
```
guiv42, line 586: warning, value -1 is used for "From the beginning" -> this leads to a crash

PR #318, opened 2024-04-02
Remove VST plugin (#312)

helge17:


helge17:
Merging manually, this gets to complicated....

Issue #269, opened 2024-03-09
rationalize definition of tuning presets


guiv42:
note: this issue is precisely documented with the intention to make it accessible to a newcomer.

# description of issue
A list of 5 tuning presets is hard-coded in [MidiSongReader](https://github.com/helge17/tuxguitar/blob/169a5e14371584cacc75fcc9d093152af2bf916c/common/TuxGuitar-midi/src/org/herac/tuxguitar/io/midi/MidiSongReader.java#L699): it is used to select an appropriate tuning preset when importing a midi file. This list has been there for a while.

More [recently](https://github.com/helge17/tuxguitar/commit/ae3a36bd4d828a9f3ae7fda5ea75f8e4ca5dea2e), a configurable list of tuning presets was added in the form of a [tunings.xml](https://github.com/helge17/tuxguitar/blob/master/desktop/TuxGuitar/share/tunings/tunings.xml) file. So now, some identical tuning presets are coded at 2 different places.

Objective is to cleanup code: remove duplicated definitions to ease future maintenance.
In a first step, actual behavior should remain unchanged. Then, tuning preset selection when importing midi files could be improved just by editing the xml.

# how to fix
Proposal: make the priority of tuning presets configurable in xml file. This is the way I would implement it:
- add an optional attribute `priority` to tuning presets in tuning.xml file. Define value 1 for standard E-tuning, 2 for Dropped-D, etc, to reproduce priorities currently hard-coded  in MidiSongReader.
- Add an integer attribute `priority` in [TGTuning](https://github.com/helge17/tuxguitar/blob/master/common/TuxGuitar-lib/src/org/herac/tuxguitar/song/models/TGTuning.java) class, with value imported from xml file content (and a default value if no attribute in xml). Import of tuning attributes from xml is done in [TuningReader](https://github.com/helge17/tuxguitar/blob/169a5e14371584cacc75fcc9d093152af2bf916c/common/TuxGuitar-lib/src/org/herac/tuxguitar/song/helpers/tuning/xml/TuningReader.java#L58)
- implement `Comparable<TGTuning>` interface in `TGTuning` class, so  instances can be sorted by `priority` order (instances with no priority attribute should come last)
[...]

PR #319, opened 2024-04-03
Configurable tuning presets in xml file using priority attributes

tdan:
modified:   common/TuxGuitar-lib/src/org/herac/tuxguitar/song/helpers/tuning/TuningManager.java
	modified:   common/TuxGuitar-lib/src/org/herac/tuxguitar/song/helpers/tuning/xml/TuningReader.java
	modified:   common/TuxGuitar-lib/src/org/herac/tuxguitar/song/models/TGTuning.java
	modified:   common/TuxGuitar-midi/src/org/herac/tuxguitar/io/midi/MidiSongReader.java
	modified:   desktop/TuxGuitar/share/tunings/tunings.xml

### DIFF
--- a/common/TuxGuitar-midi/src/org/herac/tuxguitar/io/midi/MidiSongReader.java
+++ b/common/TuxGuitar-midi/src/org/herac/tuxguitar/io/midi/MidiSongReader.java
@@ -18,6 +18,7 @@ import org.herac.tuxguitar.io.midi.base.MidiSequence;
 import org.herac.tuxguitar.io.midi.base.MidiTrack;
 import org.herac.tuxguitar.player.base.MidiControllers;
 import org.herac.tuxguitar.song.factory.TGFactory;
+import org.herac.tuxguitar.song.helpers.tuning.TuningManager;
 import org.herac.tuxguitar.song.managers.TGSongManager;
 import org.herac.tuxguitar.song.models.TGBeat;
 import org.herac.tuxguitar.song.models.TGChannel;
@@ -32,6 +33,8 @@ import org.herac.tuxguitar.song.models.TGString;
 import org.herac.tuxguitar.song.models.TGTempo;
 import org.herac.tuxguitar.song.models.TGTimeSignature;
 import org.herac.tuxguitar.song.models.TGTrack;
+import org.herac.tuxguitar.song.models.TGTuning;
+import org.herac.tuxguitar.util.TGContext;
 
 public class MidiSongReader extends MidiFileFormat implements TGSongReader {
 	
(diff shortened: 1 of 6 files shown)


common/TuxGuitar-midi/src/org/herac/tuxguitar/io/midi/MidiSongReader.java
```diff
@@ -698,45 +701,20 @@ public void checkValue(int value){
 		
 		public List<TGString> getStrings(int maxFret) {
 			List<TGString> strings = new ArrayList<TGString>();
+
+			TuningManager tuningManager = TuningManager.getInstance(new TGContext());
```
guiv42, line 705: Here, you are redefining a new instance of TuningManager, because you're recreating it from a *new context*.
This is why you needed to modify TuningManager, and to redefine the path to the resource (see remark in TuningManager file). With this blank context, the resource manager is missing information to find the resource...
I had not anticipated this. How to retrieve the correct context:

In MidiSongReader, add a property `private TGContext context`
Set this property in constructor:
```
	public MidiSongReader(TGContext context) {
		super();
		this.context = context;
	}
```

When retrieving TuningManager instance, use `this.context` as a parameter instead of `new TGContext()`

In class `MidiSongReaderPlugin`, add the `context` parameter when creating instance of MidiSongReader:
```
	protected TGSongReader createInputStream(TGContext context) throws TGPluginException {
		return new MidiSongReader(context);
	}
```

guiv42 (changes requested):
Still a few remarks to consider, one more iteration should be enough.
You can just push a new commit in your PR-ready branch
#269

common/TuxGuitar-midi/src/org/herac/tuxguitar/io/midi/MidiSongReader.java
tdan: Hey this is awesome! Yeah, I didn't know how to retrieve the correct context in the first place so that was my work-around to get the program run.

By the way, I noticed `tuningManager` would get called and had to load and process the same xml resource on every track. Would it be better to add a property `private TuningManager tuningManager` in MidiSongReader, and initialize it only once in the constructor?
 
`
public TrackTuningHelper(int track) { 
	this.track = track;
	this.maxValue = -1;
	this.minValue = -1;
	this.tuningManager = TuningManager.getInstance(MidiSongReader.this.context);
}
`

`
List<TGTuning> tunings = this.tuningManager.getPriorityTgTunings();
`

guiv42: > I noticed tuningManager would get called and had to load and process the same xml resource on every track.

It's because you're not providing the correct context.

>  Would it be better to add a property private TuningManager tuningManager in MidiSongReader, and initialize it only once in the constructor?

No. Just provide the correct context, then you will see that TuningManager.loadTunings is only called once.

PR #324, opened 2024-04-05
wider caret, for better readability

guiv42:
Evolution requested by a user on a french-speaking [guitar forum](https://forum-guitare.fr/viewtopic.php?t=58851&start=53)

@helge17: what do you think of this?
It's always a bit subjective when it comes to graphical user interface. Screenshots before and after:
original wider

Implementation details:
frustrating: cannot use painter.setLineWidth(2.0f) This would modify a property of the painter, which is used further to draw other items

Not possible either to restore lineWidth to default (or to previous value) after drawing caret: no available methods for that in UIPainter. Adding such methods in this interface is quite complex, as it has many implementations

### DIFF
--- a/desktop/TuxGuitar/src/org/herac/tuxguitar/app/view/component/tab/Caret.java
+++ b/desktop/TuxGuitar/src/org/herac/tuxguitar/app/view/component/tab/Caret.java
@@ -170,7 +170,8 @@ public class Caret {
 					
 					painter.initPath();
 					painter.setAntialias(false);
-					painter.addRectangle(x, y, width, height);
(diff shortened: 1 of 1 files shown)


helge17:
Since all the other lines are quite thin, the wider caret somehow seems a bit too thick to me. I would rather prefer to change the color of the caret from grey to maybe black or something, that would also improve visibility. You could also apply the color of the track to the caret. But then how do you distinguish voice 1 from voice 2? It really depends on personal taste, the wider caret is OK for me too.

guiv42:
Thanks for the suggestion, caret from 2.0beta looks much better. I'll have a look there.

Issue #118, opened 2023-09-25
Erroneous handling of tuplets (invalid measure duration)


guiv42:
Seen on 1.6.0, but most probably present since a long time (edit: present in 1.5.4)

The scenario below is really not critical and most users will never see this problem (time division 10:8 is not so frequent).
However, I have already seen weird behaviors of TuxGuitar regarding tuplets (triolets, etc), that I could not always understand neither reproduce. And this issue (fully reproducible) shows clearly something is going wrong under the hood.

The following sequence leads to an invalid measure (start from default template):
1. select duration 1/16h
2. select triplet, time division 9
3. select triplet, time division 10 => KO, the tuplet only contains 19 beats instead of 20, and the last beat takes an erroneous duration (1/64th)
Practically, at this stage the TGDivisionType attribute of the last beat is *set to an incorrect value*: 1/64th, no time division (should be 1/16th, time division 10:8). Possibly one loop finishing 1 step too early when setting beats duration?

What is weird is that it depends from the previous operation. The following sequence is OK (still starting from default template):
1. select duration 1/16h
2. select triplet, time division 10 -> OK

This one is OK also
1. select duration 1/16h
2. select triplet, time division 11
3. select triplet, time division 10 -> OK

https://github.com/helge17/tuxguitar/assets/129443524/ec449883-512e-4297-8035-65f5fe6107ee

Makoha:
I have encountered some unexpected behavior which I assume is also related to this issue.

If you create a full bar of 13-tupleted 64'th nodes, TuxGuitar will display the bar as expected.
![billede](https://github.com/helge17/tuxguitar/assets/73967125/0187db7d-debe-47a5-a4a6-86edbc1b8f11)

If you then go ahead and save the file and reopen it, an extra dotted 64'th node will be add to the bar.
![billede](https://github.com/helge17/tuxguitar/assets/73967125/1b668870-9fc5-449f-826a-f76931ec94d1)

I initially encountered this problem a couple a years ago using TuxGuitar 1.5.4, when I was doing an absurd project where a snare drum had to be played by some very specific intervals. Something like hitting every 7'th 13-tupleded 32'nd node.
I managed to create full bars of various tuplets and insert the notes at the intended positions, but after saving the file and reopening it, I realized that some of the bars now contained these invalid measure durations.

Makoha:
Yesterday I managed to setup Eclipse and run/build the source code.
I am currently looking into this issue.

Makoha:
[...]

PR #331, opened 2024-04-07
Improve tuplet handling

Makoha:
Improvements related to issue #118

This should allow tuxguitar to handle tuplets a lot better.
Please note that because the value of `TGDuration.QUARTER_TIME` is 960, tuxguitar is not able to compute precise durations for 7, 9, 11 and 13 tuplets. Idealy `TGDuration.QUARTER_TIME` should be set to 2882880 in order for tuxguitar to compute precise durations for all these tuplets, but simply changing `TGDuration.QUARTER_TIME` currently has some unintended consequences that would need to be handled first.
Because of this precision lose I have also added some patches that would fix some basic rounding errors when `TGDuration.QUARTER_TIME` is 960.

I am new to GitHub and contributing to open source projects, so constructive criticism is appreciated.


guiv42:
Thanks for the PR. I'll try to have a look, but this tuplet handling feature is not so trivial to me. (And this might take some time)
I already faced some difficulties when trying to improve tuplets representation (7e5e373241c41b066a8b0ec7cc05c34984c4c45b). In this case I tried to measure beats duration using an integer number of the shortest possible note, to avoid rounding effects.

Makoha:
No problem. These rounding errors are tricky to deal with, which is why I strongly would recommend working towards setting `TGDuration.QUARTER_TIME` to 2882880. (Or maybe even bigger to handle larger tuplets?)
My goal with this commit was to improve upon current functionality without breaking things.
Working on this, I believe I have gotten a fairly good understanding how durations are handled, so please ask if you have any questions.

guiv42:
> My goal with this commit was to improve upon current functionality without breaking things.

Understood. And thank you for this! (it's the same strategy I'm trying to apply since this fork started here)

I agree that a "correct" cleanup would require to modify this 960 value. But I admit it's quite difficult to evaluate all possible consequences.

guiv42 (approved):
Thanks a lot, good job!

PR #360, opened 2024-04-24
Shortcut: fix for Mac and add close current file

G-eos:
Activate shortcut for Close current, close all, close others

Clean default Shortcuts for Mac:
- remove Alt-F4 which does not work: Command-Q do the job (standard MacOs shortcut).
- remove Command-a to unify with others

Add Command-w (Mac) Ctrl-w for others to close the current file

Update translation accordingly, reuse file.closeXXX for action.file.closeXXX when available.


desktop/TuxGuitar/share/lang/messages_fr.properties
```diff
@@ -549,6 +549,9 @@ help.about.license=Licence
 #########################
 action.song.new=Nouveau fichier
 action.file.open=Ouvrir un fichier
+action.file.close-current=Fermer
```
guiv42, line 552: replace "Fermer" (close) by "Fermer le fichier courant" (close current file)
Justification: when user clicks in menu File/Close, it is quite obvious the term "close" is associated to "file"
But, in shortcuts configuration dialog, the label is displayed without any context. So, "Fermer" (close) does not provide any information to user about *what* will be closed.

Same remark for the 2 following lines: Fermer les autres fichiers, Fermer tous les fichiers.

For the same reason, I would prefer keeping all the other translations empty: de, es, it

guiv42 (changes requested):
Just 2 remarks:
1. a few corrections needed in translation files: fr, de, es, it (see my comment in fr file)
2. need to update also messages.properties! Else, in default configuration (English language), the shortcut settings dialog shows "action.file.close-current, action.file.close-others, action.file.close-all"

desktop/TuxGuitar/share/lang/messages_fr.properties
G-eos: Thanks, update pushed.
Sorry for the English missed !

guiv42 (approved):
OK, just one typo to fix -> I'll do it after merging

Issue #350, opened 2024-04-16
Export Audio File does not work for Percussion Channel


Makoha:
Steps to reproduce:
1. Open or create a tab that contains a Percussion Channel (DrumKit) beat.
2. `File` -> `Export` -> `Export Audio File`
3. The percussion track in the output audio file does not sound like a drum kit, but the sound will depend on the selected channel "Program"/"Bank". (Example: `Program #0` sounds like a piano)

Bug present in version 1.6.0, 1.6.2 and the latest beta release (tuxguitar-2024-04-13-master-windows-swt-x86_64)
Last seen working in version 1.5.6

guiv42:
bisect leads to this commit: 4b40135096dec1bf56d36b27390e49c97e8eab02

@helge17 : I assume this was back-ported from 2.0beta fork?
Unfortunately I cannot test with 2.0beta: trying to export an audio file throws an exception

Makoha:
Yes, bug is caused due to a small change in `MidiSequenceParser.java`

Original:
```java
	private void addChannel(MidiSequenceHelper sh, TGChannel channel, int track) {
		if((this.flags & ADD_MIXER_MESSAGES) != 0){
			int channelId = channel.getChannelId();
			long tick = getTick(TGDuration.QUARTER_TIME);
			sh.getSequence().addControlChange(tick, track, channelId, MidiControllers.VOLUME, fix(channel.getVolume()));
			sh.getSequence().addControlChange(tick, track, channelId, MidiControllers.BALANCE, fix(channel.getBalance()));
			sh.getSequence().addControlChange(tick, track, channelId, MidiControllers.CHORUS, fix(channel.getChorus()));
			sh.getSequence().addControlChange(tick, track, channelId, MidiControllers.REVERB, fix(channel.getReverb()));
			sh.getSequence().addControlChange(tick, track, channelId, MidiControllers.PHASER, fix(channel.getPhaser()));
			sh.getSequence().addControlChange(tick, track, channelId, MidiControllers.TREMOLO, fix(channel.getTremolo()));
			sh.getSequence().addControlChange(tick, track, channelId, MidiControllers.EXPRESSION, 127);

			if((this.flags & ADD_BANK_SELECT) != 0) {
				sh.getSequence().addControlChange(tick, track, channelId, MidiControllers.BANK_SELECT, fix(channel.getBank(), 0, 128));
			}
			sh.getSequence().addProgramChange(tick, track, channelId, fix(channel.getProgram()));
[...]

PR #366, opened 2024-04-27
Correctly export percussion tracks when using `Export Audio File`

Makoha:
Fix related to issue #350


Makoha:
There seems to be 3 classes that will be affected by the changes in this commit:
- TGSynthSongWriter
  - This is related to the `Export Audio File` plugin, which will correctly export percussion tracks due to this commit.
- MidiSongWriter
  - This is related to the `Export MIDI` plugin, which no longer will include `BANK_SELECT` control changes. 
  - `BANK_SELECT` control changes in exported MIDI-files does not seem to have been a thing since TuxGuitar version 1.2.
- MidiToAudioSongWriter
  - This is related to an `Export Gervill Audio File` plugin, that only seems to be accessible when debugging, probably due to usage of the `"com.sun.media.sound"` package, which is an internal API that has been inaccessible since JDK 9.

guiv42 (approved):
Thanks @Makoha for this fix, and the associated analysis.
I had already started discussion #185 about Gervill obsolescence, but I had missed the associated plugin. I'll create a dedicated issue for this (calling this plugin raises logically an exception).

PR #368, opened 2024-04-29
(draft) Github action addon

G-eos:
Add build from [GitHub Action](https://docs.github.com/en/actions) for MacOS-13 and Ubuntu.
This is a first trial, if you find it useful, please detail which platform and OS version is of interest, I will try to do my best to add them.

Action for Mac and Ubuntu is coming from INSTALL.md
Update: add FreeBSD build for trial too

Note: add build for MacOS 12 & 13. 14 fails (as we already know)

See result [here](https://github.com/G-eos/tuxguitar/actions)


guiv42:
For sure this approach has serious advantages.
Most contributors (if not all) do not have the possibility to build all flavors of TuxGuitar, so this approach could help detect possible incompatibilities introduced by a PR for example.

I think the most interesting use cases are:
- build with the native-modules: contributors working only in Eclipse most probably never build in this configuration (I never succeeded)
- build the Android and the FreeBSD versions: most contributors probably do not have set up the required environment
- build the macOS version: not accessible to all contributors
- for each platform, build both SWT and JFX variants


Then, I must admit there are also drawbacks. The most critical one I think: this is somehow redundant with the [build script](https://github.com/helge17/tuxguitar/blob/master/misc/build_tuxguitar_from_source.sh) used by @helge17 to build and release TuxGuitar in all flavors. Also with some information available in [INSTALL.md](https://github.com/helge17/tuxguitar/blob/master/INSTALL.md): typically the list of build dependencies.
So this means double maintenance whenever something needs to change in the build environment.

Ideally the build script, the INSTALL.md doc and the GitHub actions should share a maximum of possible data. For example:
- a dedicated script to download/install SWT could be separated from the main build script. It could be called from both the GitHub actions and the main build script, and referred to in INSTALL.md (and [contribute.md](https://github.com/helge17/tuxguitar/blob/master/doc/contribute.md) also!)
- same thing for JFX on FreeBSD
- build dependencies could be defined by a script, referenced by both INSTALL.md and the Github actions

Also: most of TuxGuitar's code is independent from the OS, and there is an abstraction layer for ui frameworks (SWT vs. JFX). So it is clearly possible to create an incompatibility, but this should really not happen very often.
After about 1 year of active development, I admit I faced personally all use cases of incompatibilities I mentioned above. So I can't say it's useless. But to be honest it happened to me about 3-4 times. Considering the number of commits I pushed, effective coverage remains quite low. Most of the incompatibilities I've introduced would not have been detected by these Github actions.

@helge17: what do you think?

helge17:
My development environment is currently quite stable and a migration to Github actions would mean a lot of effort. That's why I prefer to stick with my build script. There is a certain redundancy in the script, the documentation and the Github actions, that's true. But again, I fear that the effort to fix this redundancy is currently greater than the benefit. I also think it is better to list the individual commands in the documentation instead of just saying "call script xyz".

But if Github actions would help you, we can merge PR #368. I will try to keep the files up to date. Maybe I'll fall back on it at some point, for example if I can no longer manage to keep all my virtual machines running.

helge17:
As Github actions help you, @G-eos and @guiv42 in building TuxGuitar, I will merge this PR into master. I'll stick with my build environment for the time being, maybe I'll use the actions at some point later.

PR #376, opened 2024-05-13
add app version in user config file

guiv42:
yet unused: no functional consequence, for future use
targeted use case: to handle smooth transitions between app versions when user upgrades/downgrades app

example: would have been useful for 3190ae77b9ff3f186cbfc76a0bbdf7fe615f58b4 to adapt default behavior when user upgrades from an old version


guiv42:
@helge17 what is your opinion about this?
I'm trying to anticipate possible evolution of user config file. We already have added the possibility to create new parameters and to deprecate old ones.
With this, it should also be possible in the future to modify an existing parameter.

helge17:
I think it's a good idea: it's a simple change that can be of great benefit in the future, thank you!

PR #396, opened 2024-05-27
Separated default gmChannels for gpx

Lukramancer:
Recently I've found some `.gp` files which work fine in GuitarPro but don't in TuxGuitar. Instruments of all tracks were replaced by one of the first track.

I've discovered that they are in GP7 format without specified general midi channels at `MidiConnection` element in `score.gpif` file in them. 

empty tags

In that case TuxGuitar reads default value `0` as midi channels. The same midi channel in all tracks causes all channels but the first one skip their initialization.

I made sure that the default channel value was one that had not yet been used.

### DIFF
--- a/common/TuxGuitar-gpx/src/org/herac/tuxguitar/io/gpx/GPXDocumentReader.java
+++ b/common/TuxGuitar-gpx/src/org/herac/tuxguitar/io/gpx/GPXDocumentReader.java
@@ -122,14 +122,14 @@ public class GPXDocumentReader {
 						Node gmNode = getChildNode(trackNode, "GeneralMidi");
 						if( gmNode != null ){
 							track.setGmProgram(getChildNodeIntegerContent(gmNode, "Program"));
-							track.setGmChannel1(getChildNodeIntegerContent(gmNode, "PrimaryChannel"));
-							track.setGmChannel2(getChildNodeIntegerContent(gmNode, "SecondaryChannel"));
+							track.setGmChannel1(getChildNodeIntegerContent(gmNode, "PrimaryChannel", this.getFreeGmChannel(null)));
+							track.setGmChannel2(getChildNodeIntegerContent(gmNode, "SecondaryChannel", this.getFreeGmChannel(track)));
 						}
 					} else if (this.version == GP7) {
 						Node midiConnectionNode = getChildNode(trackNode, "MidiConnection");
 						if( midiConnectionNode != null ){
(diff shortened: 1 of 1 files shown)


common/TuxGuitar-gpx/src/org/herac/tuxguitar/io/gpx/GPXDocumentReader.java
```diff
@@ -570,4 +570,20 @@ private int[] getChildNodeIntegerContentArray(Node node, String name , String re
...
+
+		boolean isGmChannelUsed;
+		do {
+			gmChannel += 1;
+
+			isGmChannelUsed = trackToCheck != null && ((trackToCheck.getGmChannel1() == gmChannel) || (trackToCheck.getGmChannel2() == gmChannel));
+			for (GPXTrack track : this.gpxDocument.getTracks()) {
+				isGmChannelUsed |= ((track.getGmChannel1() == gmChannel) || (track.getGmChannel2() == gmChannel));
+			}
+		} while (isGmChannelUsed);
```
guiv42, line 585: Just one question: do you think it could be useful to add a criterion on max value to end this loop?
I'm definitely no expert in GM, but what would happen if returned channel number exceeds 16?

guiv42 (commented):
thanks for this PR
Just one question about possible max value for channel number.

guiv42:
I tested with a 14-track .gp song, after manual deletion of the content of all `PrimaryChannel` and `SecondaryChannel` nodes:
- current behavior
  - all tracks are allocated to Grand Acoustic Piano
- with this PR
  - tracks are allocated somehow randomly to instruments
  - only the first 8 tracks are audible
  - the list of instruments defined in song is correct: user *can* re-allocate correctly the instruments to tracks
- with this PR + a limitation of number of channels to 16
  - tracks are allocated somehow randomly to instruments
  - tracks >= 8 are all allocated to the same instrument
  - all tracks are audible
  - **but** the list of instruments defined in song is incomplete, so it's more complex for user to re-allocate correctly the instruments

So, forget my comment and let's keep this PR unchanged.

Lukramancer:
I'm not expert in GM at all. The thing about maximum channels amount is my oversight.
I'll explore that issue more, cause I figured out one more thing. Using default secondary GM channel as 0 can double max instruments available.
Can you please share that .gp file for future testing?

guiv42:
> Can you please share that .gp file for future testing?

Sure. I don't have so many .gp files, and only 1 with enough tracks, you can find it [here](https://tabs.ultimate-guitar.com/tab/dire-straits/money-for-nothing-guitar-pro-2549100)

PR #412, opened 2024-06-09
Should be: Polski

Gootector:
Should be: Polski. "Polska" is name of the country.

Greetings,
Gootector


Gootector:
@helge17 Please merge.

PR #415, opened 2024-06-12
Fix: typos

Gootector:
Fix: typos

Greetings,
Gootector


Gootector:
@helge17, @guiv42 What do you think about these changes?

helge17:
OK, thanks.

PR #419, opened 2024-06-13
Fix: typo and more tuplet names

Gootector:
Fix: typo and more tuplet names
One more "cannot" and septuplet & nonuplet names.

Greetings,
Gootector


Gootector:
@helge17 These tuplet names are taken from MuseScore. "Nonuplet" is correct. I don't know the names of the others.

PR #420, opened 2024-06-14
Update Polish Translation (Android)

Gootector:
Update Polish Translation (Android)

Greetings,
Gootector


helge17:
Please note that all Android translations are generated from the desktop messages with the misc/messages_desktop_to_android.pl script, so some of your translations may be overwritten on the next run.

Gootector:
Thanks for the info. I copied/pasted these strings from the desktop translation.

Issue #403, opened 2024-05-29
performance issue


guiv42:
TuxGuitar 1.6.3beta1, on Linux openSuse tumbleweed

tab scrolling does not follow correctly when playing, see video:

https://github.com/helge17/tuxguitar/assets/129443524/05b9b7d8-39fd-49e5-9f76-828c60b89ea5

Even with option "Player / Highlight Played Beat" unchecked.
Edit: it's not systematic... at least it's not always as severe as in this video

guiv42:
I was expecting this issue to be a consequence of the new "highlight played beat" feature, but it seems not.
I tried to `git bisect` this issue. It's not so easy, as the observed issue is not systematic, however it led me to this commit
d4d2506602a0180196f45ab122516e97a1da57f7 `Update SWT on Linux to 4.26`

Just for testing I tried (from 1.6.3beta1) to downgrade SWT to 4.21: this solves the issue.

guiv42:
technically, problem is solved by SWT downgrade.
However I prefer to leave this issue open (bug -> enhancement). Analysis shall be more detailed : we will not be able to stick to an old version of SWT forever.

guiv42:
Starting to profile app, with the objective to understand what's going on. I ran a test scenario with both versions 1.6.3beta1 and 1.6.3, and tried to profile different parts of the application.
The culprit seems to be `org.eclipse.swt.graphics.GC.drawImage` : with this scenario, cpu time consumed by this function is
- 13788ms for 1.6.3beta1
- 2154ms for 1.6.3

This SWT function is called by SWTPainter.drawImage

guiv42:
I confirm the performance issue of `org.eclipse.swt.graphics.GC.drawImage`.
Use of `GC.drawImage` by TuxGuitar:
- when updating a measure, TuxGuitar creates an image buffer in memory
- then, it paints the content of the measure into that buffer
- when buffer is ready, it calls `GC.drawImage` to display the buffer

It looks as if `GC.drawImage` in SWT 4.26 is recomputing the buffer systematically to consider zoom level (adaptation to screen resolution?):
- with 1.6.3beta1 (SWT 4.26) it spends almost all of its (long) execution time in `org.eclipse.swt.graphics.Image.getImageDataAtCurrentZoom`
- with 1.6.3 (SWT 4.21), this function is never called. Instead, `GC.drawImage` calls `GC.drawImageInPixels`

[...]

PR #423, opened 2024-06-16
Perf optimization, update SWT for Linux to 4.26

guiv42:
See #403 

@helge17 , I would appreciate your feedback before merging this PR, especially:
- not tested on macOS
- ~~not tested on FreeBSD (I've got some issues in my VM, need to spend some time on it)~~ I could finally test on FreeBSD, it's OK. Better apply the same optimization as for Linux, as it uses the same gtk flavour of SWT
- I reverted 429c42f159f4e07a3c6dded9dabc363be2478a7f: it modified `build_tuxguitar_from_source.sh` (untested)

on macOS and Windows, behavior should be unchanged.
on Linux and FreeBSD, the only visible modification is that scrollbars are now hidden when playing.
no impact on Android.

I think we shall try to update SWT for Linux quite soon, to leave us some time to detect possible issues. Now, if someone wants to package 1.6.3 for debian 12, performance issue #403 will be present.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -37,11 +37,11 @@ In order for Asian characters to be displayed correctly, you may also need to in
 ### Download and install SWT for Linux
 
 ```sh
-$ wget https://archive.eclipse.org/eclipse/downloads/drops4/R-4.21-202109060500/swt-4.21-gtk-linux-`uname -m`.zip
-$ mkdir swt-4.21-gtk-linux-`uname -m`
-$ cd swt-4.21-gtk-linux-`uname -m`
-$ unzip ../swt-4.21-gtk-linux-`uname -m`.zip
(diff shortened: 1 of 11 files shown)


helge17:
The patch works good on all OS, I did not notice any negative side effects.

> side effect on Windows, sometimes one scrollbar is not displayed correctly when player stops

With view.hide-scrollbars-when-playing=true under Windows, the vertical scrollbar is not displayed again if the cursor jumps down one line. If you stop playback in the same line, the scrollbar is displayed.

helge17 (commented):
The patch works good on all OS, I did not notice any negative side effects.

> side effect on Windows, sometimes one scrollbar is not displayed correctly when player stops

With view.hide-scrollbars-when-playing=true under Windows, the vertical scrollbar is not displayed again if the cursor jumps down one line. If you stop playback in the same line, the scrollbar is displayed.

guiv42:
Thanks @helge17 for the test. I'll have one more try, with the objective to fix the issue on Windows, so we keep a similar behavior with all operating systems. If I succeed I'll issue a new PR, else I will merge this one.

guiv42:
superseded by #433

PR #458, opened 2024-07-08
Fixes in README.md and contribute.md

Gootector:
Fixes in README.md and contribute.md

Greetings,
Gootector


Gootector:
@helge17 What do you think about my fixes? Double spaces are unnecessary. GitHub ignores them and they are displayed as a single.

Issue #462, opened 2024-07-16
feature request: improve Save/Export


guiv42:
TuxGuitar 1.6.3 (behavior identical to 1.5.6)

## Saving file under foreign format can lead to silent data loss
Open a file with a foreign format (i.e. not a TuxGuitar native file). Edit file, then Save: TuxGuitar keeps by default the original file format. This can lead to silent data loss, if the foreign format does not handle correctly the modifications.
Problem already seen several times by users: [lyrics lost when using gp3](https://sourceforge.net/p/tuxguitar/discussion/522985/thread/d3bbda8e7b/?limit=25#c206), #447, and there are other similar old tickets in SF.
Many applications issue a warning when user tries to save a file under a foreign format

## Export under foreign format
Imagine song "A.tg" is open. File/Export under another format, let's say "B.xx"

The usual behavior of an *Export* function is:
- save file "B.xx" to disk
- keep active file "A.tg"

But TuxGuitar does:
- save file "B.xx" to disk
- set "B.xx" as the active file

I find this very confusing: link with original file "A.tg" is lost. Following call to "Save" function opens "Save As". To restore the link with the original "A.tg" song it is required to overwrite it.

I'll try to understand how this works, while preparing future file format evolution

guiv42:
todo:
- [ ] save under foreign format, warn user
- [x] export, keep current file active (done, #467)

PR #470, opened 2024-07-21
warn user when saving under foreign format

guiv42:
@helge17: I would like your opinion about this PR before merging

The idea is to avoid silent data loss (#462), **but** it conflicts with your [recommendation](https://github.com/helge17/tuxguitar/discussions/41#discussioncomment-8572347) to issue a warning when we change file format.
We won't be able to issue 2 inconsistent warnings:
- warning, latest file format is not backward compatible, use with care
- warning, please use latest file format or you may loose data

In any case, these 2 sentences will be relevant, and it will be difficult for user to choose.

Possible way out: use new extension ".tg2" when we change file format, and rewrite warning message created in this PR.

### DIFF
--- a/common/TuxGuitar-lib/src/org/herac/tuxguitar/io/base/TGFileFormatManager.java
+++ b/common/TuxGuitar-lib/src/org/herac/tuxguitar/io/base/TGFileFormatManager.java
@@ -8,6 +8,7 @@ import org.herac.tuxguitar.event.TGEventManager;
 import org.herac.tuxguitar.io.tg.TGFileFormatDetectorImpl;
 import org.herac.tuxguitar.io.tg.TGSongReaderImpl;
 import org.herac.tuxguitar.io.tg.TGSongWriterImpl;
+import org.herac.tuxguitar.io.tg.TGStream;
 import org.herac.tuxguitar.util.TGContext;
 import org.herac.tuxguitar.util.singleton.TGSingletonFactory;
 import org.herac.tuxguitar.util.singleton.TGSingletonUtil;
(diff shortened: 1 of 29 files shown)


helge17:
I would suggest keeping the tg extension for all TuxGuitar file versions and the following behavior:

- The "Save" and the "Save as" dialog only allow to save the file in the latest tg version.
- "Save" is executed immediately only if the opened file has already been saved in the latest tg format.
- All other file formats (older tg formats, gp3, midi etc.) are only imported, i.e. if the user clicks on "Save" (or "Save as") after the import, the "Save as" dialog is displayed, which only allows saving in the latest tg format. If the user then explicitly decides to overwrite an existing tg file, a corresponding warning is displayed (sometihng like „File exists, do you really want to overwrite the file with the latest tg file version?). 
- All formats except the latest tg format can only be exported. After the export, TuxGuitar remains on the original tg file (or the previously imported file if the file has not yet been saved in the latest tg version).
- The "Open" dialog should still be able to open all supported file formats, but should behave like "Import" for all formats except the latest tg format. BTW, the dialog contains separate entries for all TuxGuitar versions. As all formats use the same extension tg, I think it would be better if there was only one "TuxGuitar" entry in the list.

With the current patch #470, I also noticed that when opening a tg3 file, the file is immediately set to zero size when you click on "Save".

helge17:
You are right, it is better not to show the warning about updating the tg file when saving in newer versions. After all, everyone should update to the latest TuxGuitar version sooner or later. The problem with 2.0beta was that it was kind of a dead end.

But I would keep the extension tg and not use tg2 or similar. Guitar Pro went through gtp, gp3, gp4, gp5, gpx and has now arrived at gp, I wouldn't want that for TuxGuitar.

guiv42:
superseded by #481

PR #481, opened 2024-07-31
Foreign formats

guiv42:
See #462, and [this post](https://github.com/helge17/tuxguitar/pull/470#issuecomment-2246236013) in #470

@helge17: this should be close to your proposal. I would appreciate some more testing if you can.


helge17:
The PR #481 is working fine, thanks! I think the last open point is a clear definition of „File → Open“ and „File → Import“: Currently some formats are "opened", others "imported", and I think it is not clear to the user what exactly the difference is. I see two options:

1. remove "Import" altogether and just use "Open" (and move TEF & MIDI to "Open"). Also on Android there is no "Import", only "Open".
2. use "Open" only for *.tg, all other formats will be imported.

I'm tending towards 1, but maybe there are good arguments for 2.

guiv42:
Good
If that's OK for you, I suggest merging #481 as is, then I'll have a look at "Open" vs. "Import". I agree with you, option 1. seems more logical.
Yet there is a difference between Open and Import:
- File/Import one .midi file, it shows a dialog enabling user to transpose song
- File/Open, select "All Files" in file formats filter (else .midi is not visible), then select a .midi file, file is opened without showing the dialog

PR #483, opened 2024-08-02
move ".tef" from File/Import to File/Open

guiv42:
see [this comment](https://github.com/helge17/tuxguitar/pull/481#issuecomment-2262278439) in PR481

This PR moves ".tef" files from File/Import to File/Open menu. I am clearly hesitating for midi files.
Looking at all file formats TuxGuitar can read: TablEdit, PowerTab, GuitarPro and TuxGuitar all produce *tablature* files. Basically, the same *nature* of data encoded differently. For these, I think File/Open makes sense.
Midi files contain data of a fundamentally different *nature*. The imported data is significantly transformed when TG opens a midi file. So I think it makes sense to keep it in File/Import menu. It makes more clear to users the data is "imported", hence "modified".

By the way, I have to fix something: following 340f89993c8a9206ff4030c87ed4fc37d1faefa3, old TG file formats don't appear any more in File/Open (as expected), but they are now visible in File/Import (unexpected).

User doc also needs to be updated

### DIFF
--- a/common/TuxGuitar-lib/src/org/herac/tuxguitar/io/base/TGFileFormatManager.java
+++ b/common/TuxGuitar-lib/src/org/herac/tuxguitar/io/base/TGFileFormatManager.java
@@ -23,7 +23,7 @@ public class TGFileFormatManager {
 	
 	private List<TGFileFormatDetector> fileFormatDetectors;
 	private List<TGFileFormat> commonReadFileFormats;
-	private TGFileFormat nativeWriteFileFormat;
+	private TGFileFormat nativeFileFormat;
 	
 	private TGFileFormatManager(TGContext context){
 		this.context = context;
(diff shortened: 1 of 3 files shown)


guiv42:
old TG file formats should no more appear in File/Import

I am currently considering a more serious rework of Open/Import and file formats management. Implementation in this PR should work I think, but it looks a bit quick & dirty.

@helge17 : if it's OK for you I think it can be merged, by I would prefer to take some more time before issuing a 1.6.4

helge17:
That's perfectly fine with me, just merge the PR when you're ready. I just tested it very quickly and it looks pretty good.

guiv42:
OK, thanks.
Just changed my mind: if a global rework of file format manager seems useful, it may be more appropriate to do this together with file format evolution. So, not in 1.6.4, which remains a minor version.

PR #505, opened 2024-09-06
Update Polish Translation

Gootector:
Update Polish Translation

Greetings,
Gootector


android/TuxGuitar-android/res/values-b+pl/strings.xml
```diff
@@ -57,7 +57,7 @@
 	<string name="action_view_layout_show_chord_name">Pokaż nazwy akordów</string>
 	<string name="action_view_layout_show_chord_diagram">Pokaż diagramy akordów</string>
 
-	<string name="action_transport_play">Odtwarzaj/zapauzuj utwór</string>
+	<string name="action_transport_play">Odtwarzaj/wstrzymaj utwór</string>
```
guiv42, line 60: @Gootector : please don't modify android `string.xml` file. These files are generated from the files in `common/resources` before release.
See here:
https://github.com/helge17/tuxguitar/blob/647c7826acbfa11ad0a551480192f52815b1b8a6/android/TuxGuitar-android/res/values-b%2Bpl/strings.xml#L5

guiv42 (changes requested):
see my comment: android translation files shall not be modified manually

Gootector:
> see my comment: android translation files shall not be modified manually

If it is as you typed here, it will be generated automatically. My changes will be replaced (or not) and I don't have to do anything with this PR. Please merge my PR. Thank you.

guiv42:
Well, that's right, I could have merged it...
Thanks for the update

Gootector:
> Well, that's right, I could have merged it... Thanks for the update

https://github.com/helge17/tuxguitar/commits/master/android/TuxGuitar-android/res/values-b%2Bpl - As you can see, the last changes (PR) before 1.6.4 release are made by me. No "generate from the files in common/resources before release"?

guiv42:
Hmm, that's annoying.
The script to generate android translations from "common" files is [there](https://github.com/helge17/tuxguitar/blob/master/misc/messages_desktop_to_android.pl). The problem is that it's not automatized yet, so it's possible to forget this step before releasing.
I don't know what's best here. It seems difficult to me to ask all translators to run this perl script by themselves. We've got to think at a better solution.

helge17:
The build script now checks if the Android translations are up to date when releasing a new stable version (8af678c32f238a1c4ad9e553cd03ca67cc5161b7).

Issue #533, opened 2024-09-25
feature request: improve main toolbar


guiv42:
I am currently considering a partial redesign of the main toolbar. The objective is to provide a significantly visible improvement for a future 2.0 version.
To all users: if you have suggestions for improvement, that's the right moment.

My *objectives* (don't know if I will succeed):
- integrate completely the player dialog in the main toolbar, and delete the dialog. Almost all controls present in this dialog are already accessible in the main toolbar.
  To add in toolbar:
  - metronome button
  - currently playing tempo
  - numeric display, with a timestamp instead of current midi tick counter
- enable display/hide fretboard and instruments in one single click (replace drop-down menu by buttons)
- make content of toolbar more configurable (e.g. possibility to display/hide some sections)

This is what it could look like (Breeze skin):
![toolbar](https://github.com/user-attachments/assets/ca7bb261-5b3d-497b-b44b-76559be82f22)

belfie13:
I think the customizable part should be the focus, everything should have a single click icon that each user can customize because we all use different features. I would even like a switchable option so i can change between toolbars depending on my workflow. but this seems like a big task?

nobody459:
i think the main problem the current toolbar has is that the icons too small.
also the button to skip to the next section is painful to use. otherwise it s fine.

guiv42:
@nobody459 : this is probably configuration-dependent (no problem at all on my side). Could you please precise your config?
- screen size
- screen resolution
- operating system

nobody459:
> @nobody459 : this is probably configuration-dependent (no problem at all on my side). Could you please precise your config?
> 
> * screen size
> * screen resolution
> * operating system

around 30 inch
HD something
Intel Core i7

guiv42:
Well... these are answers, but not to my questions!

- screen resolution: I need the precise values, how many pixels width / height
- operating system: Windows 10? Windows 11? MacOS (and which version)? Linux? ... ?

nobody459:
sorry !

regarding the system i use windows 11.

regarding the pixels i dont know but it s an Acer Aspire 512.

guiv42:
screen resolution should be accessible with Windows display settings. Something like a right-click on the desktop should lead you there

guiv42:
> I think the customizable part should be the focus

@belfie13 could you please elaborate? What precisely would you like to be able to configure?

You can test the current implementation of this new toolbar in tuxguitar-next branch (see pre-releases).
[...]

Issue #355, opened 2024-04-18
MacOS Cocoa plugin shows error with SWT 4.15 and above


helge17:
When upgrading SWT to version 4.15 or higher, the MacOS Cocoa integration plugin shows the following error at startup:

![MacOS14_SWT_cocoa-plugin](https://github.com/helge17/tuxguitar/assets/5897631/3b4aa009-8bdd-4473-a3d0-5f63109d6fa4)

```
MacOS14 00-Binary_Packages % ./tuxguitar-2024-02-22-snapshot-macosx-swt-cocoa-x86_64.app/Contents/MacOS/tuxguitar.sh
org.herac.tuxguitar.util.plugin.TGPluginException: org.eclipse.swt.internal.cocoa.NSWindow.standardWindowButton(long)
	at org.herac.tuxguitar.cocoa.toolbar.MacToolbarPlugin.connect(MacToolbarPlugin.java:28)
	at org.herac.tuxguitar.util.plugin.TGPluginManager.connectPlugin(TGPluginManager.java:88)
	at org.herac.tuxguitar.util.plugin.TGPluginManager.connectEnabled(TGPluginManager.java:65)
	at org.herac.tuxguitar.app.TuxGuitar.startUIContext(TuxGuitar.java:141)
	at org.herac.tuxguitar.app.TuxGuitar.access$100(TuxGuitar.java:69)
	at org.herac.tuxguitar.app.TuxGuitar$1$1.run(TuxGuitar.java:124)
	at org.herac.tuxguitar.app.synchronizer.TGSynchronizerControllerImpl$1$1.run(TGSynchronizerControllerImpl.java:29)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:185)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4262)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3885)
	at org.herac.tuxguitar.ui.swt.SWTApplication.start(SWTApplication.java:64)
	at org.herac.tuxguitar.app.TuxGuitar.createUIContext(TuxGuitar.java:117)
	at org.herac.tuxguitar.app.TuxGuitar.createApplication(TuxGuitar.java:102)
	at org.herac.tuxguitar.app.TGMainSingleton.launchTuxGuitar(TGMainSingleton.java:57)
[...]

PR #544, opened 2024-10-06
main toolbar rework (integrate player dialog)

guiv42:
see #533
@helge17: I would like to have your opinion before merging, it's always highly subjective when it comes to user interface.
This evolution targets a version 2.0 and is inspired by 2.0beta fork, but I'm following a more conservative path.

Timestamp display is the most challenging evolution:
- font is configurable, just click on timestamp
- most probably a specific setting will be needed for font on macOS, could you please have a look?
- foreground & background colors are configured for some skins (see skin.prop files)
- timestamp zero is always displayed when not playing, it's intentional. 2.0beta always displays a timestamp, but when not playing the value becomes erroneous after the first "repeat close" (could not try when playing, I can't make it work). Parsing the whole song is necessary to compute a correct timestamp, it's done when player starts but cannot be done permanently.
- I hope it will work correctly on macOS, I have struggled to get something working correctly in all configs [Linux,Windows]*[SWT,JFX]


helge17:
@guiv42, I just cloned your toolbar branch. It seems to work fine on BSD and Linux and for me it looks good!

Also the JFX version works on MacOS (maybe I have to fine tune the font settings a little bit):
toolbar_jfx

But the SWT version freezes on MacoS with the following error:
toolbar_swt_error

```
% ./tuxguitar-2024-11-04-toolbar-macosx-swt-cocoa-x86_64.app/Contents/MacOS/tuxguitar.sh
java.lang.NullPointerException: Cannot read field "id" because the return value of "org.eclipse.swt.widgets.Control.paintView()" is null
	at org.eclipse.swt.widgets.Control.drawWidget(Control.java:1265)
	at org.eclipse.swt.widgets.Widget.drawRect(Widget.java:776)
	at org.eclipse.swt.widgets.Display.windowProc(Display.java:6043)
	at org.eclipse.swt.internal.cocoa.OS.objc_msgSendSuper(Native Method)
	at org.eclipse.swt.widgets.Display.applicationNextEventMatchingMask(Display.java:5284)
	at org.eclipse.swt.widgets.Display.applicationProc(Display.java:5716)
	at org.eclipse.swt.internal.cocoa.OS.objc_msgSend(Native Method)
	at org.eclipse.swt.internal.cocoa.NSApplication.nextEventMatchingMask(NSApplication.java:97)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3804)
	at org.herac.tuxguitar.ui.swt.SWTApplication.start(SWTApplication.java:64)
	at org.herac.tuxguitar.app.TuxGuitar.createUIContext(TuxGuitar.java:122)
	at org.herac.tuxguitar.app.TuxGuitar.createApplication(TuxGuitar.java:103)
	at org.herac.tuxguitar.app.TGMainSingleton.launchTuxGuitar(TGMainSingleton.java:57)
	at org.herac.tuxguitar.app.TGMainSingleton.launchSingleton(TGMainSingleton.java:45)
	at org.herac.tuxguitar.app.TGMainSingleton.main(TGMainSingleton.java:35)
```
Unfortunately, the MacOS code is a little outdated.

I also noticed that when playing a selection, the timestamp always starts at zero. Is that intentional?

guiv42:
Thanks for your feedback.

That's bad news for macOS: I can't debug this, and it seems the problem comes from SWT itself. I will try to downgrade SWT on Linux to see if I can reproduce, but I'm not very optimistic.
Result is a bit weird: several items are not drawn correctly (timestamp canvas, quarter icon for current tempo, icons for metronome and countdown), but tempo value and icon for playing mode are OK.

> I also noticed that when playing a selection, the timestamp always starts at zero. Is that intentional?

I've seen this also. To be honest, no it wasn't intentional. But finally I think it makes sense. If you think it's better to display the absolute timestamp I can have a look.
I was also wondering: when not playing, it seems too complex to display a correct timestamp, especially because of repeats. In this case, what do you think is better: display "0:00:00.0" (as currently implemented), or just don't display anything?

guiv42:
I've had a very quick look at macOS/Linux SWT source codes, they are too different I will not reproduce.
So I think we have no choice but to upgrade SWT (#355)
@helge17 : could you please apply your [patch](https://github.com/helge17/tuxguitar/issues/355#issue-2251153045), build with SWT4.26, and see what the toolbar looks like?
I'll try to refine analysis of #355.

guiv42:
> Your toolbar branch works fine with SWT 4.26 on macOS!

Good, thanks!

> Should I apply the patch [#355 (comment)](https://github.com/helge17/tuxguitar/issues/355#issue-2251153045) to tuxguitar-next? It would also help me to upgrade my build system from macOS 11 to 14.

No, not yet. I agree we shall target this SWT upgrade, but it's a bit early. I'm currently analyzing #355, more details to come soon. 

> Maybe the patch [#355 (comment)](https://github.com/helge17/tuxguitar/issues/355#issuecomment-2077452622) from 2.0beta helps to update the macOS code.

Unfortunately I don't think so.
 
> I think a black display is confusing, so I would prefer "0:00:00.0", or better "-:--:--.-"?

You're right, "-:--:--.-" looks the best solution, I'll do this and update PR. Thanks!

guiv42:
@helge17 : thank you for merging #571, it's not every day we succeed to tackle a macOS-specific issue :)
If you don't mind, I will also let you merge this one, as it deserves a minimum testing in macOS configuration (logically it should be OK)

Issue #579, opened 2024-11-22
Feature Request: Allow Measures to be Invalid Duration While Editing


j-t-m-e:
Request for flexible bar duration, or to allow a bar to be too long or too short while editing.

As soon as I enter a note it automatically fills the rest of the bar with rests. This makes it difficult to edit the duration of notes after the fact because I find myself struggling with the strict limitation of not letting the bar be anything but exactly full.

Forcing all measures to be permanently valid greatly simplifies the job from software point of view, but allowing measures to be invalid duration would make editing and composing easier because you can under-fill or over-fill a measure temporarily while adjusting the duration of any note.

belfie13:
yes this is a great feature i think. constantly running into problems where shifting one note will shift the entire score.

i'm not sure how much is involved with playing measures out of length but i'd be happy even if the measure is marked as incorrect length and it won't play.

guiv42:
Yes, this is for sure an interesting option I'll investigate. But it's really complex, so it will take time.

@j-t-m-e just a question: I clearly understand the added value of this tolerance, but what I don't understand is why filling a measure with rests is a problem. Basically, if I want to create new notes, I need first to place the caret on an existing beat (e.g. a rest), then only I can add my note(s). For example, if at some point in time I have a 4/4 measure with one single quarter in it (temporarily invalid measure), how could I add new beats?

guiv42:
In fact, it *is* currently possible to define invalid measures, but it's a *bug* (e.g. see #118). Note that it doesn't prevent song to be played or saved.
Before going any further on this feature request, an unambiguous evaluation of measure validity is definitely required: #118 needs to be fixed.

belfie13:
> For example, if at some point in time I have a 4/4 measure with one single quarter in it (temporarily invalid measure), how could I add new beats?

either insert new beat or when navigating right over a bar line of a measure with less beats than the time signature it will automatically add a rest.

so if you have one single quarter in a 4/4 and your cursor is on the note, when you press right to go to the next measure it will add a rest instead.

and could add an action to fill the invalid measure with rests.

guiv42:
I've started the development of this feature.
Basically the idea is to add an optional "free edition mode", disabled by default so that users are not confused. In this mode, some actions would be performed without any control. Identified so far:
- insert beat
- delete beat
- change note duration
- change measure time signature

I'm probably missing others. Please complete this list if you can think of any other action.

[...]

PR #591, opened 2024-12-03
fix tuplets rounding errors

guiv42:
see #118


Makoha:
Hi @guiv42 

I see that your pull request is commits to `tuxguitar-next`, which is a branch I haven't managed to build yet.
At the moment I get a lot of errors which seems to be caused by the fact that packages in `TuxGuitar-lib` have changed from `org.herac.tuxguitar` to `main.java.org.herac.tuxguitar`.

Do you have any wise words how to fix this?

guiv42:
> I see that your pull request is commits to tuxguitar-next

Yes, there's a good reason for that. In tuxguitar-next we changed the `.tg` file format. It's a good opportunity to store `TGBeat.preciseStart` in saved files instead of `TGBeat.start`, to fix the rounding errors (not yet implemented, I was waiting for your review)

> At the moment I get a lot of errors which seems to be caused by the fact that packages in TuxGuitar-lib have changed from org.herac.tuxguitar to main.java.org.herac.tuxguitar.

I changed this to introduce automated unit testing. What errors do you get?
I can think of 2 possible problems:
1. unit tests fail if your absolute path to tuxguitar folder contains non-ascii characters. If needed, unit tests can be bypassed by adding `-DskipTests` to your maven build command (e.g. `mvn clean verify -DskipTests`)
2. Eclipse cannot switch properly between master and tuxguitar-next branches, because its hidden `.project` files are excluded from git repo.

What I can suggest:
- clone my "preciseDurationLong" branch in a dedicated folder, different from your usual working folder, to avoid conflicts with `.project` files. Or you may experience problems when coming back from tuxguitar-next to master... (I had to struggle about this before understanding where the problem was)
- completely re-import everything into an independent Eclipse project. The procedure is detailed in [contribure.md](https://github.com/helge17/tuxguitar/blob/master/doc/contribute.md)

Makoha:
I accept your concern about changing `TGDuration.QUARTER_TIME`. I just want to still be able to experiment.
I'll look more into your pull request, but these where just my initial notes.

guiv42:
OK, thanks, that's clear.
793d388a61c455854bc44d5e454997456daefc05 fixes the sextuplet issue.

About my PR, the general idea is this one: we cannot delete `TGBeat.start` attribute, and we shall assume it is imprecise, just because of the past. So, add on top of it another `preciseStart` attribute, without rounding errors.
In the future, the ideal solution would be:
- when parsing an input file, set `TGBeat.start` attribute for each beat (no choice - compatibility with older versions)
- after full file is read,  and whatever the format, update all `preciseStart` attributes in the song (easy to implement)
- then, *never* manipulate `TGBeat.start`

Ideally `TGBeat.setStart()` method should only be called from file readers. It's a long way to go before reaching this point... until then, we have to consider this new `preciseStart` *may not always be available*: whenever it is needed for some time computation, update corresponding values for all beats in a measure.

guiv42:
`TGBeat.preciseStart` attribute is now saved to new file format. Warning: this changes the file format in an incompatible manner.
I merged tuxguitar-next branch back here to solve the conflicts associated to test files

guiv42:
@Makoha : could you have a look at this PR?
Just tell me if you think you don't have time for this.

Makoha:
Hi @guiv42
Sorry for the long delay, I have been pretty hung up lately.
I just managed to test an old .tg file (v1.5) that has a lot of weird tuplet changes, that previous versions of tuxguitar was not able to correctly playback. But your ”preciseDurationLong” branch is now able to play it correctly, which is really great.

New issue I have encountered is that if you make a new bar, select 64’th notes and then select ”Dotted” or ”Double-Dotted”, you still get an invalid bar.
I know this is an extreme example, but so was some of the tuplet cases that wanted to be fixed.

guiv42:
Thanks @Makoha for your feedback.
I know the possible sub-divisions shorter than a 64th are not completely managed by my current implementation. But to be honest I was expecting that trying to set a dotted (or double-dotted) 64th would rather fail than create an invalid bar.
I will try to debug that.

The key function is [`TGDuration.splitPreciseDuration`](https://github.com/guiv42/tuxguitar/blob/d9054713d7ad868a8969ac8f1af88254deddb440/common/TuxGuitar-lib/src/main/java/org/herac/tuxguitar/song/models/TGDuration.java#L141): how to split an interval into a list of valid TGDuration instances.

guiv42:
OKay... I had a quick look, and the result is quite suprising.

Description of test:
- open TG with default template
- click on 64th -> all rests in bar are converted to 64th
- click on "dotted"

This gives the surprising result below


What is even more surprising: I am NOT convinced this is an invalid measure (?)
I'm currently working in parallel on another feature in a dev branch, to authorize measures to be invalid during edition (#579), and for this I implemented a new feature: use red color for invalid bars.
Here is how an invalid bar created by TG 1.6.6 is displayed:

This one IS definitely invalid.

I haven't done the computation by hand... but since the measure in first screenshot is not displayed in red, it might as well be valid (?)

Anyway, I think the only clean solution to this issue is the one suggested in #579: allow measures to be invalid during edition.

So, if it's OK for you I will merge this PR as it is currently.

guiv42:
@Makoha: could you please have a look now? Hopefully this should be fixed.

I initially thought I could avoid to multiply by 4 the duration of a whole for this dotted/double-dotted issue, but finally I had to do it, to avoid rounding issues. That changes a bit the `splitPreciseDuration` method. I tried to comment the code quite heavily, because it's not trivial.
Basically, I do exactly the same as before, but 3 times successively for each possible time division:
- try to split only with no dotted duration
- if it fails, try to split with dotted
- if it fails, try to split with double-dotted

That should be ok now

https://github.com/user-attachments/assets/a99c3f47-3261-43d4-a489-44735979f437

Makoha:
Seems good, I haven't been able to find any issues so far.
Would your approach also work if we added more tuplets, like 17-tuplets and 19-tuplets?

guiv42:
Yes that would work. The possible divisions of a whole's duration is [calculated dynamically](https://github.com/guiv42/tuxguitar/blob/6e4ece1a01be80ffdae3a8afc780771e7be94796/common/TuxGuitar-lib/src/main/java/org/herac/tuxguitar/song/models/TGDuration.java#L79) from the list of division types.
But that would break once again the file format: if we publish this PR in TG 2.0 and add new time divisions later, then version 2.0 would not be able to open the new files.

Makoha:
Would there be away to avoid breaking the 2.0 file format when adding new time divisions?
Why does the new file format need to save the duration? Couldn't it just be relatively computed based on the actual note values used in the "song"?

guiv42:
> Would there be away to avoid breaking the 2.0 file format when adding new time divisions?

Yes, and no. Compatibility could be maintained in many cases by creating a new xml node for the new duration, and by saving both the old and new duration attributes to the file.
But what if a new file effectively contains a beat with a 17:16 time division? The old version would definitely be unable to interpret it.
That would make more sense to break compatibility explicitly I think.

> Why does the new file format need to save the duration?

Well... I'd say that's the current data model structure, that I don't want to break. Beat start timestamp is attached to `TGBeat` (logical), but duration is not: it's attached to `TGVoice`.

guiv42:
My suggestion: try to keep it simple (or at least not too complex!)
So, either keep the PR as it is, or immediately add the ability to handle 17-tuplets. So that there's only 1 file format evolution breaking compatibility.
I must admit the first option is my preferred one!
If you think adding 17-tuplets really has an added value, could you eventually test this configuration? Just add it in `TGDivisionType`, then build TG with `mvn clean verify -DskipTests` command (or else the unit tests will fail: some test .tg files would need to be updated)

Makoha:
Keeping it as it is is probably the best decision for now.
I'll try to do some more tests this weekend, but I feel this PR is pretty stable.

guiv42:
Hi @Makoha : could you perform some additional tests?
If it's OK for you, I will merge this PR as it is in the coming days. Thanks for your support!

guiv42:
One bug fixed: having increased so much the values to encode preciseStart could lead to an overflow (property was erroneously read as Int, and correctly written as Long)
That should be OK now

Issue #289, opened 2024-03-16
Missing symbols and notation in MusicXML export.


Skippern:
I use TuxGuitar mainly on Linux but also on Windows to view and edit GP tabs, and export to MusicXML for use in Rocksmith+ Workshop. I notice that a lot of the notation is missing in the exported MusicXML.

Missing these symbols makes the process of importing songs to Rocksmith+ very cumbersome.

Below list is an observation of missing symbols and notations, compiled with help of different users of Rocksmith+ Workshop:

Tuning:
* [ ] Tuning-Alter (sharps / flats in tuning)
* [ ] Capo

Bends:
* [ ] Bend
* [ ] Bend + Release
* [ ] PreBend
* [ ] PreBend + Bend
* [ ] PreBend + Release

Repeats:
* [ ] Repeats
* [ ] Alternative Endings

Note statuses:
* [ ] Start Tie
* [ ] Ghost Note
* [ ] Accented Note
* [ ] Heavily accented Note
* [ ] Staccato
* [ ] Palm Mute
* [ ] Dead note (FHM)
* [ ] Let Ring
* [ ] Natural harmonics
* [ ] Artificial harmonics
* [ ] Pinch harmonics - technically not supported in Tux.

Slides:
* [ ] Legato Slide
* [ ] Shift Slide
* [ ] Slide in from below
* [ ] Slide in from above
* [ ] Slide out downward
* [ ] Slide out upwards

Techniques:
* [ ] Hammer on
* [ ] Pull off
* [ ] Tap
* [ ] Slap
* [ ] Pop

Picking technique:
* [ ] Tremolo
* [ ] Trill

guiv42:
Hi,

which version of TuxGuitar are you using? We are currently working on musicXML export, so it is constantly evolving.
In latest snapshot, tuning-alter, repeats and alternative endings should be OK. Lyrics are in progress (#282). Drums parts are still buggy (I'm currently working to fix this).

Then, you're right there's still a lot to do.

However it won't be possible to export everything, e.g. because of limitations in musicXML format itself. I haven't looked in details at all your list, but for example the description of bend effect available in [musicXML](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/bend/) is not detailed enough to export all information available in TuxGuitar.
Some other fields might be quite tricky, because the internal representation of data in TuxGuitar really differs from the musicXML data structure (thinking of "hammer-on/pull-off" for example).

Skippern:
[...]

PR #596, opened 2024-12-06
Improvements to MusicXML exporter, and some changes to the GPX importer.

ffio1:
This change is to address some issues called out in #289 . When you would import a GPX file, and export to MusicXML, and compare the resulting MusicXML file with one provided by other software, or even what TuxGuitar could read from the GPX file, there was a lot of things missing in the export. This change attempts to address that with the following changes.

GPX import changes:
- Importing Capo / Offset
- Importing Let Ring
- Importing Slap
- Importing Pop
- Fixes an issue where the trill fret was imported as a midi note, instead of fret, causing the resulting data to be incorrect.
- Adds importing of TGDuration to trills, where-as it was previously left as default.

UI changes:
- Adds support for quarter note, and eighth note trills as they are supported on imports, but it is impossible to see that right now as we only show 16, 32, and 64th note trills in the trill dialog.

MusicXML exporter changes:
- Bug fix for repeats (wrong amount of repeats was previously written)
- Added Repeat alternate endings
- Added Capo / Offset
- Added Dynamics
- Added Markers / Rehearsals
- Added Natural / Artificial Harmonics
- Added Ghost Notes
- Added Slides
- Added Palm Mutes / Dead Notes
- Added Accents / Heavily Accented Notes
- Added Staccato
- Added Hammer-on / Pull-off
- Added Tapped Notes
- Added Bends / Pre-bends / Releases
- Added Tremolos
- Added Trills


desktop/TuxGuitar/src/org/herac/tuxguitar/app/view/dialog/trill/TGTrillDialog.java
```diff
@@ -33,6 +33,8 @@ public class TGTrillDialog {
 	private UIRadioButton sixtyFourthButton;
 	private UIRadioButton thirtySecondButton;
 	private UIRadioButton sixTeenthButton;
+	private UIRadioButton eighthButton;
```
guiv42, line 36: This is not fully consistent with actual .tg file format:
- create a new song, add a note and associate an eighth trill
- save file under .tg format
- open this file with TuxGuitar 1.5.x or 1.6.y
- try to edit the trill: only the 3 original radio buttons are displayed, no one is checked. Click OK: this deletes the trill (quite counter-intuitive)

I can suggest 2 solutions:
1. postpone the merge of this PR *after* the evolution of .tg file format (implemented in tuxguitar-next branch)
2. (my preference): revert this modification of TGTrillDialog, and modify `GPXNote.setTrillDuration` to make sure longest possible duration for an imported trill is a sixteenth, whatever the real duration was in GP file. Musically, I don't really understand what an eighth or quarter trill would be.

@ffio1 : what is your opinion?

guiv42 (commented):
Thank you @ffio1 for this PR!

I just have one remark/question about the maximum possible duration of a trill (see my comment in the code).

Whatever option we choose for this open point, this PR will need to wait the publication of release 1.6.5 before being merged (it's a bit too late to be integrated in 1.6.5, beta is already out)

desktop/TuxGuitar/src/org/herac/tuxguitar/app/view/dialog/trill/TGTrillDialog.java
ffio1: @guiv42 I think it would be best to preserve the data if at all possible, so it may be best to wait. I'm in no rush to get this in, so whenever you think it's best to do this is fine with me.

For me to better understand the issue you're saying, if someone makes a tab on my branch with one of these new trill options, then tries to import it into an OLDER version of TuxGuitar before my changes, it will show none of them checked, and clicking OK removes the trill that existed in the data, but wasn't visualized? Is that correct?

The reason for this addition is I was making GP files in Guitar Pro, and moving them to TuxGuitar, and comparing both GP and Tux's MusicXML. In GP's UI, there are options for quarter, eighth, sixteenth, and sixty fourth, so it was a bit odd that half the options that would get exported weren't being imported (or rather... they were, but it was never visible).

image

If you think it's best not to support those trill cases, I can also remove this. This is not a major thing for me, it was just to make sure we were visualizing the data we were importing once I fixed the trill duration not being imported.

guiv42: > For me to better understand the issue you're saying, if someone makes a tab on my branch with one of these new trill options, then tries to import it into an OLDER version of TuxGuitar before my changes, it will show none of them checked, and clicking OK removes the trill that existed in the data, but wasn't visualized? Is that correct?

Yes, that's perfectly correct. It's not critical, but the job to redesign a new .tg file format has already been done (in tuxguitar-next branch), so this is a good opportunity to avoid this inconsistency.

> I think it would be best to preserve the data if at all possible

OK, fine for me. So let's just wait for tuxguitar-next branch to be merged into master before merging this PR.
By the way, I will modify this new file format to limit the trill duration to 1 quarter max (currently there's no limit)

Issue #471, opened 2024-07-22
Missing hammer-on/pull offs in TEF files


mattrobineau:
The hammer-on and pull-offs are not being added when opening a TEF file in TuxGuitar.
TablEdit:
![image](https://github.com/user-attachments/assets/5301c695-1d08-4769-ba49-fea1140e730e)

TuxGuitar:
![image](https://github.com/user-attachments/assets/a19122b6-7afc-4f1e-9493-63e22fb4e23e)

guiv42:
@mattrobineau: could you please provide a tef file so I can reproduce the problem?
I tried to download TablEdit demo version, but for an unknown reason TuxGuitar cannot open files I generate from this app.
Ideally I would like 3 different tef files, as simple as possible:
- reference file : 1 measure with a few notes, no hammer-on, no pull-off
- hammer-on file: same file as the reference, with one hammer-on
- pull-off file: same file as the reference, with one pull-off

mattrobineau:
Unfortunately I also only have the demo version and cannot get those tefs to import into TuxGuitar.

For now I guess I can just manually apply the missing things.

guiv42:
OK
Until someone who can access TablEdit provides such test files I don't think I can progress. As far as I know tef format is closed and not documented.

Makoha:
I did a bit of research on TablEdit, and it seems like a new file format was released alongside the release of V.3.00, which was release somewhere around the year 2020.
I guess that TuxGuitar is not able to read this new format, but there should be a function in TablEdit that allows the user to save as v2 format.
However, this function is not available in the demo version.

ffio1:
I've never used TablEdit before, but could be a good first issue. I'll take a look at this.

I was able to find an old v2 installer on archive (dot) org so I could make test arrangements. It appears a lot of techniques were missing in the TuxGuitar import, as well as dynamics, so I'll include those in my changes.

PR #597, opened 2024-12-07
#471 Adds better technique importing for v2 TEF files

ffio1:
Addresses an issue when importing v2 TEF files where techniques, and dynamics were not being imported.

Changes include:
- Dynamics
- Hammer On
- Pull Off
- Natural Harmonic
- Artificial Harmonic
- Palm Mute
- Tapping
- Vibrato
- Tremolo (only 1 duration though. No metadata for duration)
- Bend / Bend + Release
- Let Ring
- Slap
- Ghost Note
- Staccato
- Fade In

### DIFF
--- a/desktop/TuxGuitar-tef/src/org/herac/tuxguitar/io/tef/TEInputStream.java
+++ b/desktop/TuxGuitar-tef/src/org/herac/tuxguitar/io/tef/TEInputStream.java
@@ -225,13 +225,14 @@ public class TEInputStream {
 			
 			if( ((data[2] & 0xff) & 0x1f) > 0  && ((data[2] & 0xff) & 0x1f) <= 25 ){
 				int duration = (data[3] & 0xf);
-				int dynamic =  (data[3] >> 4);
-				int effect = data[4];
+				int dynamic =  ((data[3] & 0xff) >> 4);
+				int effect1 = data[4];
+				int effect2 = data[5];
 				int fret = (((data[2] & 0xff) & 0x1f) - 1);
 				if((((data[2] & 0xff) >> 5) & 0x01) != 0 ) {
 					fret += (data[5] & 0xff);
 				}
-				this.song.getComponents().add( new TEComponentNote(position, measure, string ,fret,duration,dynamic,effect ) );
+				this.song.getComponents().add( new TEComponentNote(position, measure, string ,fret,duration,dynamic,effect1,effect2 ) );
 			}
 			else if( ((data[2] & 0xff) & 0x1f) == 27 ){
 				//TIME SIGNATURE CHANGE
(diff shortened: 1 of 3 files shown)


desktop/TuxGuitar-tef/src/org/herac/tuxguitar/io/tef/TEInputStream.java
```diff
@@ -225,13 +225,14 @@ private void readComponents(){
 			
 			if( ((data[2] & 0xff) & 0x1f) > 0  && ((data[2] & 0xff) & 0x1f) <= 25 ){
 				int duration = (data[3] & 0xf);
```
guiv42, line 227: A problem not directly related to this PR: I was surprised by the encoding of `dynamic` field, as its least significant bit was unused (see `TESongParser.getVelocityFromDynamic`), so I tried to force this bit manually.

It appears this bit corresponds to "double-dotted" attribute. So:
- decoding of `dynamic` is OK, it's the 3 most significant bits of `data[3]`
- but `duration` is encoded on the **5** least significant bits of `data[3]` (and not 4 as currently defined), with most significant bit == "double-dotted"

@ffio1 : do you think you could fix this?

guiv42 (commented):
Thank you very much @ffio1 for this PR.
I followed your idea to get an old version of TablEdit for a few tests. I did not test exhaustively your PR, but I just detected an error about decoding of one note's duration (see my remark in the code). Note that this error was there already before this PR.
OK for me to merge this, after 1.6.5 has been released

desktop/TuxGuitar-tef/src/org/herac/tuxguitar/io/tef/TEInputStream.java
ffio1: @guiv42 Yeah I can fix that!

ffio1: @guiv42 I was able to get my v3 stuff moved to the side temporarily, and committed a fix for this. I did not see the same MSB for double dotted that you were seeing.

What I was seeing that 19, 22, 25, and 28 were the double dotted variants. While I was there, I also fixed the whole note dotted case (31) as it was being imported with the wrong duration.

guiv42: > I did not see the same MSB for double dotted that you were seeing

I did not test as precisely as you did: I was just surprised by this unused bit, tried to set it manually to 1 for one note, and saw a double-dotted note appear in TablEdit. I did not go any further in the analysis.

Thanks!

Issue #500, opened 2024-08-28
TGActionException opening tef files


rschader:
I have a few tef files I recently downloaded, and when I try to open them, I get an error:
org.herac.tuxguitar.action.TGActionException. Some of the other tef files I downloaded from the same site open up fine. Maybe the tef files are corrupt? They were zipped up, no errors in the zip files. I've attached the 2 tef files that cause this error.
[Crazy.zip](https://github.com/user-attachments/files/16784229/Crazy.zip)
[I Never Cried.zip](https://github.com/user-attachments/files/16784232/I.Never.Cried.zip)

rschader:
I was able to open the problem files fine using the Tefviewer software, so it must be an issue with Tuxguitar, or maybe something wrong with my computer?

guiv42:
@rschader : there is no problem with your computer, I confirm the issue (TuxGuitar 1.6.4, Linux openSuse tumbleweed).

I did a basic binary comparison with some .tef files I have on my machine, structure of these file is quite different. Most probably there are several versions of TablEdit out there, producing different file structures with the same .tef extension (as TuxGuitar does).

The difficult part is that, as far as I know, the tef file format is closed and proprietary.

guiv42:
I have contacted TablEdit development team. It is confirmed the .tef file format changed in 2020, so no big surprise if TuxGuitar cannot read the new format.
TablEdit team is currently not willing to share the definition of this file format publicly. That is fully understandable for a commercial software.
So unless someone has a TablEdit license and some time to reverse-engineer the new format, this issue will not progress.

helge17:
According to the [TablEdit changelog](https://tabledit.com/welcome/whatsn_e.shtml) the file format changed from V.2.78 to  V.3.00. Commit ea3af745b116c6ad309082fcdba61fb3a5931f38 adds a "v2" to the documentation.

ffio1:
I'm working on a change to add TablEdit v3 support. Will post an update in a few days, hopefully with a PR.

guiv42:
I have good news.
After some discussions with TablEdit editor, I do have a personal access to a description of tef v3 file format. I cannot share it publicly, but I have an explicit authorization to use it for an implementation of tef v3 decoding in TuxGuitar.

@ffio1 : what is the current status of #600? Did you go any further in the implementation?
Do you want me to take the lead on this feature?

ffio1:
[...]

PR #600, opened 2024-12-14
TablEdit v3 import support

ffio1:
Addresses #500 where version 3 of TablEdit changed the file format used. This is such a large change I believe it is necessary to split this into two separate plugins like we do with GTP / GPX as parsing these files only has a few similarities like note duration.

Please let me know if you have any feedback on this, or find anything that may not have been accounted for already. I tried getting everything I could find, but I haven't used TablEdit that much so I could have very well missed something as there are still some unknown pieces to the file format.


desktop/TuxGuitar-tef3/src/org/herac/tuxguitar/io/tef/TEFileFormatDetector.java
```diff
@@ -0,0 +1,41 @@
...
+				return TESongReader.FILE_FORMAT;
+			}
+			return null;
+		} catch (Throwable throwable) {
+			return null;
+		}
+	}
+
+    public boolean isSupportedVersion(int majorVersion, int minorVersion, int revision) {
+        return majorVersion == 3;
```
guiv42, line 39: This is very risky: any file with byte number 3 equal to 0x03 will be identified as a tef v3 file!
Better check all the 4 first bytes.

desktop/TuxGuitar-tef3/src/org/herac/tuxguitar/io/tef/TESongReader.java
```diff
@@ -0,0 +1,48 @@
...
+	
+	public void read(TGSongReaderHandle handle) throws TGFileFormatException {
+		try {
+			TESong teSong = new TEInputStream(handle.getInputStream()).readSong();
+			TGSong tgSong = new TESongParser(handle.getFactory()).parseSong(teSong);
+			
+			handle.setSong(tgSong);
+		// } catch (Exception e) {
+		// 	throw new TGFileFormatException(e);
+		// }
```
guiv42, line 30: please remove outdated comments

guiv42 (changes requested):
Thanks for the PR.
I haven't done a full review: especially I did not really look at the detailed implementation of tef v3 decoding. That will probably need some significant amount of time to test everything. Support of the community will be needed there: I think it can be published in a pre-release.

I have a few concerns, mainly:
- separation of problems: it would be much better if this PR only dealt with tef v3. This PR cannot be merged as is: it's raising merge conflicts
- is it really necessary to make 2 different java modules? I think 1 single java module could host 2 different plugins
- implementation of file format detection needs to be improved I think

and another minor point: you can also edit AUTHORS file ;)

guiv42:
See discussion in issue #500

guiv42:
PR closed, superseded by following PRs:
- tef v2: #622 
- tef v3: #623

Issue #585, opened 2024-11-27
move one beat left does not work


guiv42:
configuration: TuxGuitar 1.6.4, Linux openSuse tumbleweed
bug is reproduced with TuxGuitar 1.5.6

video (1.5.6) is self-explanatory

https://github.com/user-attachments/assets/92a6df63-2e75-4618-a2d3-0b2af314cce9

ffio1:
Proposed fix sent in PR #601.

guiv42:
Before going any further let me ask a trivial but important question: what are "move beat" actions *supposed to do*? It's not that obvious.
Current behavior is measure-wide and seems to be:
- move right: insert one rest (with duration of current note) and shift right all following beats of measure. This can only be done if measure ends with enough rest(s), which are then deleted
- move left: delete current beat, shift all following beats left and insert rest at measure end

Important note: there are not so many controls available in Android ui, but these 2 actions are present and as far as I know nobody complained. Does this mean this behavior is acceptable? No idea.

belfie13:
sounds like move right should be called beat insert rest and move left should be called beat remove note!

guiv42:
> sounds like move right should be called beat insert rest and move left should be called beat remove note!

Yep, it looks like that. But that doesn't make sense as it collides with other items available in the same menu: Beat -> Insert Rest Beat, Beat -> Delete Note or Rest.

I've got to check the implementation

guiv42:
I've written a few erroneous statements above, so let's stick to the code.
Forget Android: I've been confused by the "left/right arrow" icons, these are completely unrelated to "move left/right".

# move left action
- deletes current beat
- moves all following beats *of track* left by the duration of the selected *voice*

Behavior of menu item "Beat -> Delete Note or Rest" is quite different:
- if selected voice contains notes, it deletes the notes
- else, if another voice contains notes, it does nothing
- else if no voice contains note
  - if selected voice is the longest one it does nothing
  - else it removes the beat, moves all following beats *of measure* left, and inserts a rest at measure end

# move right action
I's not so clear. It always inserts a rest beat, with a duration corresponding to selected *voice*. But depending from beat content, this rest may be inserted *before* or *after* the beat (!)

https://github.com/user-attachments/assets/0eaa5a55-f061-45d6-96bb-3f5010bea30e


To sum up: I'm lost.

guiv42:
[...]

PR #601, opened 2024-12-15
Fix Move one Beat left button

ffio1:
Addresses #585 .

Original issue: When moving to the left, the `theMove` parameter is negative, and would go down the path of removing beats. With this, we would end up removing the beats that had the notes and would not re-add them.

Fix: Adds a hash-map of <time after move, list of notes pre-move> so we can keep track of what notes need to be moved, and add them to the new (post-move) beats.

### DIFF
--- a/common/TuxGuitar-lib/src/org/herac/tuxguitar/song/managers/TGTrackManager.java
+++ b/common/TuxGuitar-lib/src/org/herac/tuxguitar/song/managers/TGTrackManager.java
@@ -2,6 +2,7 @@ package org.herac.tuxguitar.song.managers;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


guiv42:
I just performed a few tests, it does not solve the issue in all cases, see video

https://github.com/user-attachments/assets/a42f2c90-ce9b-4cea-898b-ed945c28437e

guiv42 (commented):
see my comment in the PR discussion, this does not solve completely the issue

ffio1:
> I just performed a few tests, it does not solve the issue in all cases, see video

Thanks for catching that! I was blindly copying the notes from the deleted beat's voices to the new beat voice's. This didn't take the new voice's duration into consideration, and it also had the side effect of other beat metadata being removed (chord, text event, etc.). I will try to address this.

Issue #472, opened 2024-07-22
Repeat Endings or First and second endings not showing for TEF files


mattrobineau:
The repeat endings shown below do not show up when opening TEF files in TuxGuitar.

TablEdit:
![image](https://github.com/user-attachments/assets/df91432b-0758-4773-86e9-1a6d032ccb60)

TuxGuitar:
![image](https://github.com/user-attachments/assets/40cce9f6-2de1-4181-998a-6c527090fce6)

ffio1:
Alternate endings will be added to TablEdit v2 import support alongside #600.

guiv42:
Should be OK now
PR #600 superseded by PR #622 for tef v2 endings

guiv42:
@mattrobineau : can this issue be closed?
You can test latest [pre-release](https://github.com/helge17/tuxguitar/releases), just take care to use a version with suffix "master"

mattrobineau:
@guiv42 I'm still not seeing it. I used this release: tuxguitar-2025-01-19-master-linux-swt-amd64.tar.gz 


![Image](https://github.com/user-attachments/assets/ead9bd08-6d6e-498e-b6a2-1b1c2ce6a3df)

The orange highlight of measure 8 and 9 are missing the first ending and 10 is missing the second.

guiv42:
OK
Could you please provide a test file? (just zip it and attach it here in this conversation)

mattrobineau:
[endings.zip](https://github.com/user-attachments/files/18624433/endings.zip)

Contains 2 files, endings.tef was created using TablEdit v3.04 a4c demo version, the other is from an unknown version (to me) and is the one I've used in the screenshots..

guiv42:
OK thanks. For the 2nd file, if you could open it with tuxguitar-2025-01-19-master then it's most probably a tef v2 file.
I'll have a look

guiv42:
Here is what I get when I open the tef v3 file with TuxGuitar (in tuxguitar-next branch), is that correct?

![Image](https://github.com/user-attachments/assets/e14ead17-4027-4d4c-bb9c-e904158bbdd5)

mattrobineau:
Yes, looks almost identical except repeat is x2 and pauses are a little different. But in regards to the issue with endings, it looks good.

![Image](https://github.com/user-attachments/assets/5d4c43dc-4b2c-4637-a39f-af82e2a0b58d)

guiv42:
[...]

PR #622, opened 2024-12-30
Add tempo change / ending / repeat TEF v2 support

guiv42:
Adds support for tempo changes, alternative endings, and repeats components in the TEF v2 import process.

This PR re-uses implementation done initially in PR #600
(splitting tef v2 and tef v3 into different PRs)


guiv42:
should fix issue #472

PR #623, opened 2024-12-30
add TablEdit v3 support

guiv42:
This PR should fix issue #500, it re-uses tef v3 implementation done in PR #600, with the following modifications:
- tef v2 and v3 are 2 different plugins, but in the same java module
- minor modification of tef v3 file header decoding (essentially file format check)
- modification of file format detection:
  - more precise identification of tef v3 by file content 
  - ".tef" file extension is only associated to tef v2, as tef v3 files should be identified by file content (and not file extension)

@ffio1 : I would appreciate if you could review this PR
As you will see, I did not take the time to deeply review the detailed tef v3 decoding, it's exactly your initial implementation.


desktop/TuxGuitar-tef/src/org/herac/tuxguitar/io/tef3/TEInputStream.java
```diff
@@ -0,0 +1,1132 @@
...
+		if (256 != this.readBytes(headerBytes)) {
+			return false;
+		}
+		byte majorVersion = headerBytes[3];
+		// unused: minorVersion = headerBytes[4], subVersion = headerBytes 5 and 6
+		if (3 != majorVersion) {
+			return false;
+		}
+		for (int i=16; i<24; i++) {
+			if (0 != headerBytes[i]) {
```
ffio1, line 133: Just curious, is this something you got from the spec?

I wasn't able to trigger these bytes in Tabledit, so I also think these should be null, but do we need to check that they're 0 because they're part of the version check, or is it something we could live without as it could be something that we just haven't seen yet.

ffio1:
Code looks good to me. Just a question on one of the header checks, if it's something you cannot answer that's OK too.

desktop/TuxGuitar-tef/src/org/herac/tuxguitar/io/tef3/TEInputStream.java
guiv42: > Just curious, is this something you got from the spec?

Yes, this information comes from TablEdit editor. This corresponds to a kind of "protected mode", only accessible with the paid version. This explains why you could not set these bytes to any other value than zero.

guiv42:
OK, thanks for you review

Issue #641, opened 2025-01-22
conflict between bends and tuplets


guiv42:
TuxGuitar 1.6.6, Linux (openSuse tumbleweed).
When score is hidden and only tablature is displayed, bends and tuplets graphical representations conflict.

That's because effects are displayed *above* the tuplets (see vibrato on screenshot)

![Image](https://github.com/user-attachments/assets/d14a3e7e-4b85-4ecf-86a8-41b70fb7c4ab)

guiv42:
This one will require some significant work to be fixed.
It's a big mess in [`TGTrackSpacing`](https://github.com/helge17/tuxguitar/blob/6ef2a18685424ee1bfc87fca6cf1161ea40ff5cd/common/TuxGuitar-lib/src/main/java/org/herac/tuxguitar/graphics/control/TGTrackSpacing.java): the values defined in the tables do not match with the comments next to them.
Note: it was already the case in version 1.5.6 when this fork started.

It seems these table are just completely wrong. I tried to re-order the different items to be aligned with the effective rendering, this creates significant spacing issues.
My interpretation: vertical spacing has been adapted manually (and empirically) at several places in the code to make things look nice, despite the incorrect configuration tables.
Need to fix all this before even trying to change anything. Not trivial at all

Issue #562, opened 2024-10-24
Feature request: ’Effects'  No marking for the direction of plucking strings on electric guitar


sbfksmq:
![gh1](https://github.com/user-attachments/assets/6983a520-fa4b-4d37-81dd-d6bf0f00c191)
![gh2](https://github.com/user-attachments/assets/acc23b45-bff6-45d9-9e73-f74ba2d95e1d)
Feature request: ’Effects'  No marking for the direction of plucking strings on electric guitar

guiv42:
It seems one implementation was done back in 2022: https://sourceforge.net/p/tuxguitar/patches/46/
To be analyzed.
The new file format (currently implemented in tuxguitar-next branch) could support this feature. Graphical conflict with bends needs to be fixed.

guiv42:
I'm currently trying to merge the implementation from SF.
A few graphical conflicts need to be fixed:
- conflict with pre-bend/release
- conflict with tuplets when score is masked

![Image](https://github.com/user-attachments/assets/415d4b99-8260-40b9-9e0e-e9a12e01019f)

guiv42:
Implemented

guiv42:
available in pre-release of tuxguitar-next branch (>=2025-02-03-tuxguitar-next)

PR #642, opened 2025-01-23
Pick stroke

guiv42:
In progress, implementation of pick stroke direction (#562)

Original implementation: https://sourceforge.net/p/tuxguitar/patches/46/
Done so far:
- merge the original patch
- adapt to the new file format
- fix graphical conflict with tuplets when score is hidden

Still TODO:
- icons: original implementation in SF provides 2 icons, in png format. Need to create vector graphic versions, and to handle light/dark skins
- I failed to implement it in Android app. Android Studio did not like at all the re-organization of folders I have done in tuxguitar-next branch (addition of a folder "main" in the path) to enable unit testing when building desktop app. It displays hundreds of errors everywhere, because packages names do not match with path.
- refer to discussion in SF (link above): it seems this original implementation creates issues with GP (not yet analyzed)

I don't want to fix the graphical conflict with bends (#562) in this PR: bends also conflict with other elements (#641).
To be handled in an independent PR (need to rework bends vertical position, no direct impact on pick stroke)


guiv42:
I could finally fix the issue on Android.
Android Studio still *displays* a lot of errors, but it *builds* app correctly.

@helge17 : could you take care of the icons?
Also, I'm not quite sure about the way the Android translations are re-used from common conf files. I've tried to run [misc/messages_desktop_to_android.pl](https://github.com/helge17/tuxguitar/blob/master/misc/messages_desktop_to_android.pl) but without success. So I ended up editing the android xml files manually.

helge17:
Commit b59d0be1a3fade4575bde63de8037ef4d5d9ddf0 adds the pick stroke icons. I will care about the translations as soon as this PR is merged.

May Android build shows 25 warnings, most of them due to deprecation.

guiv42:
Thanks @helge17 
Do you think this PR can be merged now? You have mentioned on SF that the initial patch caused a few problems with Guitar Pro 6. Can you reproduce them?

helge17:
When opening the exported files in GP6 I haven't noticed any more problems, so I think the PR is fine, just merge it, thanks!

Issue #659, opened 2025-03-01
Android: Java error in menu on long tap on empty position


helge17:
A long tap on an empty position in the tablature shows the following error in tuxguitar-next (not in master):

![Image](https://github.com/user-attachments/assets/1c590f91-dfd8-4b4d-8499-de92d30abfc6)

The error seems to be related to commit 9d6d34b6de3a580402fe29a46ab9b67828482dae.

logcat output:
```
02-27 20:55:38.681 14658 14658 W System.err: java.lang.NullPointerException: Attempt to invoke interface method 'android.view.MenuItem android.view.MenuItem.setOnMenuItemClickListener(android.view.MenuItem$OnMenuItemClickListener)' on a null object reference
02-27 20:55:38.690 14658 14658 W System.err:    at org.herac.tuxguitar.android.menu.controller.TGMenuBase.initializeItem(TGMenuBase.java:34)
02-27 20:55:38.690 14658 14658 W System.err:    at org.herac.tuxguitar.android.menu.controller.TGMenuBase.initializeItem(TGMenuBase.java:41)
02-27 20:55:38.690 14658 14658 W System.err:    at org.herac.tuxguitar.android.menu.controller.TGMenuBase.initializeItem(TGMenuBase.java:45)
02-27 20:55:38.690 14658 14658 W System.err:    at org.herac.tuxguitar.android.menu.controller.impl.smart.TGSelectedBeatMenu.initializeItems(TGSelectedBeatMenu.java:47)
02-27 20:55:38.690 14658 14658 W System.err:    at org.herac.tuxguitar.android.menu.controller.impl.smart.TGSelectedBeatMenu.inflate(TGSelectedBeatMenu.java:31)
02-27 20:55:38.690 14658 14658 W System.err:    at org.herac.tuxguitar.android.menu.controller.TGMenuCabCallBack.onCreateActionMode(TGMenuCabCallBack.java:35)
02-27 20:55:38.690 14658 14658 W System.err:    at com.android.internal.policy.DecorView$ActionModeCallback2Wrapper.onCreateActionMode(DecorView.java:2598)
02-27 20:55:38.690 14658 14658 W System.err:    at androidx.appcompat.view.SupportActionModeWrapper$CallbackWrapper.onCreateActionMode(SupportActionModeWrapper.java:159)
[...]

PR #660, opened 2025-03-01
fix android issue, long tap on empty position

guiv42:
trying to fix issue #659 

@helge17 , could you please check if it's OK?
I've got serious problems in my AndroidStudio configuration, for an unknown reason I'm constantly loosing communication with virtual device, so it's not so easy for me to test :(

### DIFF
--- a/android/TuxGuitar-android/res/menu/menu_selected_beat.xml
+++ b/android/TuxGuitar-android/res/menu/menu_selected_beat.xml
@@ -35,5 +35,8 @@
     <item
         android:id="@+id/action_change_stroke"
         android:title="@string/action_change_stroke"/>
+    <item
+        android:id="@+id/action_change_pickstroke"
+        android:title="@string/action_change_pickstroke"/>
 
 </menu>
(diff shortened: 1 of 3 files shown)


helge17:
The patch fixes issue #659, thanks!

I noticed (with this patch) that when long tapping on an empty position, the "Pick stroke up/down" item is present in the three-dot overflow menu at the top right, while when tapping on a note, this menu item is missing.

In general, there are some differences between the desktop and Android menus, but maybe it's best if you merge this PR and I'll take a closer look afterward.

guiv42:
Thanks for your feedback.

> I noticed (with this patch) that when long tapping on an empty position, the "Pick stroke up/down" item is present in the three-dot overflow menu at the top right, while when tapping on a note, this menu item is missing.

I think I should be able to fix this. I'll have a look before merging

guiv42:
That should be OK now (added "pickStroke" item in selected note menu).
Could you please test again?

helge17:
Works good now, thanks!

Issue #680, opened 2025-03-16
feature request: import grace notes from gp files


guiv42:
TuxGuitar 1.6.6, openSuse tumbleweed

TuxGuitar currently import grace notes from .gp files as normal notes. This creates invalid measures (too long)

PR #684, opened 2025-03-18
Add GPX grace note support

ffio1:
This should address #680 

GP8:
image

Tux:
image

GP project upload: 
[GP_Project.zip](https://github.com/user-attachments/files/19324731/GP_Project.zip) | Github won't let me upload the .gp file so it's in the zip.

Unfortunately I didn't find a better way at handling this, but if you have any suggestions please let me know. GPX has the grace notes on a separate beat from the notes they are linked to, so we need to parse the grace note beat, then the real beat after and reconnect the grace notes.

Additionally GP7/8 support multiple back-to-back beats of grace notes, which I don't believe would work in TuxGuitar as grace notes are effects on regular TGNotes, and can only have 1 grace note effect. In that case, I count every second grace note beat as a real note. Here's a view in GP and then Tux, where 1(X) - 9 are grace notes, and 10 is a regular note.

GP8:
image

Tux:
image

If you want more samples from GP, let me know.

### DIFF
--- a/common/TuxGuitar-gpx/src/app/tuxguitar/io/gpx/GPXDocumentParser.java
+++ b/common/TuxGuitar-gpx/src/app/tuxguitar/io/gpx/GPXDocumentParser.java
@@ -1,5 +1,6 @@
 package app.tuxguitar.io.gpx;
 
+import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 
@@ -35,6 +36,7 @@ import app.tuxguitar.song.models.TGTrack;
 import app.tuxguitar.song.models.TGVelocities;
 import app.tuxguitar.song.models.TGVoice;
 import app.tuxguitar.song.models.effects.TGEffectBend;
+import app.tuxguitar.song.models.effects.TGEffectGrace;
 import app.tuxguitar.song.models.effects.TGEffectHarmonic;
 import app.tuxguitar.song.models.effects.TGEffectTremoloBar;
 import app.tuxguitar.song.models.effects.TGEffectTremoloPicking;
@@ -261,9 +263,38 @@ public class GPXDocumentParser {
 					GPXVoice voice = this.document.getVoice( voiceIds[v] );
 					if( voice != null ){
 						long tgStart = tgMeasure.getStart();
+
+						GPXBeat previousBeat = null;
+						List<GPXNote> previousBeatGraceNotes = null;
 						for( int b = 0 ; b < voice.getBeatIds().length ; b ++){
 							GPXBeat beat = this.document.getBeat( voice.getBeatIds()[b] );
 							GPXRhythm gpRhythm = this.document.getRhythm( beat.getRhythmId() );
+
+							// Grace notes in GPX are on a separate beat.
+							// For us, grace notes are on the same beat as the next note.
+							// So we need to read the grace GPXNotes here, but need to use them on the next beat
+							if (beat.getGraceNotes() != null)
+							{
+								// We need to ensure we only skip grace notes if the previous beat was not grace notes
+								// GPX supports multiple back-to-back beats being grace notes, while we do not.
+								// In this case, we treat the second grace note beat as a real note.
+								if (previousBeatGraceNotes == null){
+									previousBeatGraceNotes = new ArrayList<>();
+
+									for (int n = 0; n < beat.getNoteIds().length; n++){
+										GPXNote gpNote = this.document.getNote( beat.getNoteIds()[n] );
+	
+										if (gpNote == null){
+											continue;
+										}
+	
+										previousBeatGraceNotes.add(gpNote);
+									}
+	
+									previousBeat = beat;
+									continue;
+								}
+							}
 
 							TGBeat tgBeat = getBeat(tgMeasure, tgStart);
 							TGVoice tgVoice = tgBeat.getVoice( v % tgBeat.countVoices() );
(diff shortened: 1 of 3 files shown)


guiv42:
Thanks for the PR, I'll have a look.

> Additionally GP7/8 support multiple back-to-back beats of grace notes, which I don't believe would work in TuxGuitar as grace notes are effects on regular TGNotes, and can only have 1 grace note effect

That is for sure a problem. I can see several options when there are multiple grace notes:
1. import one single note as grace note, and the other ones as normal notes
2. import one single note as grace note, and discard the other ones
3. try a mix: import some notes as grace, others as normal notes

I understand you chose option 3. I would vote for option 2.
Options 1 and 3 will end up with an invalid measure (too long). In a next release of TuxGuitar this would issue a warning when opening file, open a dialog and show all invalid measures in red. Also, user will receive a warning whenever trying to save/export this invalid song. See #579 (implementation is in progress, you can test it in [tuxguitar-next](https://github.com/helge17/tuxguitar/tree/tuxguitar-next) branch, or [latest pre-release](https://github.com/helge17/tuxguitar/releases/tag/2025-03-17-tuxguitar-next))

ffio1:
For option 2, what grace note should be preserved?

In my example from 1 - 9 being grace notes, I would assume either 1 (X) or 9 but would like to know if you'd rather the first grace note in the chain, or the last grace note in the chain be preserved.

guiv42:
> In my example from 1 - 9 being grace notes, I would assume 9 as it's the closest to the real note

Yes it makes sense, at least if grace notes are "before beat".
I'm not quite sure what to do if they are "on beat". Maybe keeping the last one is also a good solution?

ffio1:
I agree keeping the last one makes sense, even in the case of "on beat".

Here's what it looks like with it only using the last note, which I believe matches what you are looking for for option 2: 
image

Here's an edge case I also saw that will result in data loss but with how grace notes are handled, I'm not sure if we can properly handle this without modifying how we store grace notes. Grace notes into a rest beat. I'm not sure if this is proper notation though, just something I saw when I was removing some grace notes and then tried to reimport into Tux.

Guitar Pro 8:
image

Imported into Tux:
image

common/TuxGuitar-gpx/src/app/tuxguitar/io/gpx/GPXDocumentParser.java
```diff
@@ -261,10 +263,43 @@ private void parseBar(GPXBar bar , TGMeasure tgMeasure){
 					GPXVoice voice = this.document.getVoice( voiceIds[v] );
 					if( voice != null ){
 						long tgStart = tgMeasure.getStart();
+
+						GPXBeat previousBeat = null;
+						List<GPXNote> previousBeatGraceNotes = null;
```
guiv42, line 268: just a question: since only the last grace note is considered, why store a list (and why not only 1 single "previousBeatGraceNote")?
my interpretation: need to store a list, because until next beat has been read, it's not clear which grace note in the list should be kept (as TG only tolerate grace note on the same string as the real note it's associated to). Is this correct?

common/TuxGuitar-gpx/src/app/tuxguitar/io/gpx/GPXDocumentParser.java
```diff
@@ -261,10 +263,43 @@ private void parseBar(GPXBar bar , TGMeasure tgMeasure){
...
 							GPXRhythm gpRhythm = this.document.getRhythm( beat.getRhythmId() );
 
+							// Grace notes in GPX are on a separate beat.
+							// For us, grace notes are on the same beat as the next note.
+							// So we need to read the grace GPXNotes here, but need to use them on the next beat
+							if (beat.getGraceNotes() != null)
+							{
+								// We need to ensure we only skip grace notes if the previous beat was not grace notes
+								// GPX supports multiple back-to-back beats being grace notes, while we do not.
+								// In this case, we treat the second grace note beat as a real note.
```
guiv42, line 281: this comment needs to be updated

guiv42:
About the corner cases (e.g. "grace notes into a rest beat") I'm aligned with the implementation of this PR: keep it as it is.
Data loss cannot be avoided as the internal data structure of TG differs significantly from this of GP.

common/TuxGuitar-gpx/src/app/tuxguitar/io/gpx/GPXDocumentParser.java
ffio1: Yes that is part of it, but there's also a piece that a grace note could be a chord and we want to make sure we get all of the notes in that grace note chord. If we only stored one note, we could end up with only 1 of the grace notes being imported when all of them should be imported.

common/TuxGuitar-gpx/src/app/tuxguitar/io/gpx/GPXDocumentParser.java
ffio1: Fixed.

guiv42:
@ffio1 : how did you apply the style guidelines? (as you mentioned in the description of commit fa7a71d612ff8904aed31a391fb0e575bd1bb3e1)

According to [contribute.md](https://github.com/helge17/tuxguitar/blob/master/doc/contribute.md), style formatting should be applied as defined by [tuxguitar_formatting.xml](https://github.com/helge17/tuxguitar/blob/master/misc/tuxguitar_formatting.xml). When I do this in Eclipse it applies tabs for indentation, not spaces

ffio1:
Hmm I'll take a look. I'm not using eclipse as I can't compile with it, instead I'm using a Linux VM, so I'm using VS Code. With that, there is a plugin to format with eclipse settings, but it may not be following it exactly.

guiv42:
I confirm the xml file defines tabs for indentation:
https://github.com/helge17/tuxguitar/blob/c939b7525edb10e9f1fb7d79f92944073911dc24/misc/tuxguitar_formatting.xml#L398

guiv42:
Let me merge this PR as it is, then I'll fix the formatting issue

PR #688, opened 2025-03-27
Update: new GitHub username + Polish translation

Pruchniakowski:
Update: new GitHub username + Polish translation

Greetings,
Grzegorz


guiv42:
Can you please also update the [user help](https://github.com/helge17/tuxguitar/blob/c939b7525edb10e9f1fb7d79f92944073911dc24/desktop/TuxGuitar/share/help/about.html#L183)?

This duplication of data is not very practical, but help files need to be raw html, while authors file is displayed here in github (expecting markdown), and in app "Help -> about" dialog (expecting raw txt)

Issue #689, opened 2025-03-29
Why we have a minimum 30bpm restriction? It should be allowed to have 1bpm.


smartacephale:

guiv42:
It's pretty easy to release the constraint, but I want to understand first.
@smartacephale : are you *really* working with less than 30bpm, or is it just a problem of tempo *base duration* (i.e. what is a *beat*)?
Just thinking about a song at 10 quarters per minute, that would probably not be very readable, and changing the base tempo should logically help. You can have a try with latest [pre-release](https://github.com/helge17/tuxguitar/releases), that enables you to change tempo base duration. E.g. you can define tempo at 40 *beats* per min, with a *beat* defined as an eighth (equivalent to 20 quarters per min), or even a sixteenth (=10 quarters per min) etc.

smartacephale:
@guiv42 It's not a problem, I just want to work with lower tempo, with 30bpm if you really want to stretch anything, you have to create million additional measures and notes and so on and so on. I think it would be super cool and useful if you lift this restriction.

guiv42:
I must admit I don't understand. But it's not a big deal, so I will implement this evolution

guiv42:
done (#692)

PR #690, opened 2025-03-31
Change minimum bpm "larghissimo" from 20 to 1.

smartacephale:
https://github.com/helge17/tuxguitar/issues/689


guiv42:
No, that's not the right way to do it. The file you modified in in the `deprecated` folder, it will not change anything.
Nevermind, I'll do the modification myself.

PR #694, opened 2025-04-03
Improve MusicXML export

rettinghaus:
This PR brings

- writes the optional capo element only when needed
- marks the tab staff as "alternate"
- adds support for let-ring elements
- marks tab notes as stemless
- minor formatting

NB: LetRing is exported as hanging tie. If this should be the "L.R." text, it has to be exported as such.

### DIFF
--- a/desktop/TuxGuitar-musicxml/src/app/tuxguitar/io/musicxml/MusicXMLWriter.java
+++ b/desktop/TuxGuitar-musicxml/src/app/tuxguitar/io/musicxml/MusicXMLWriter.java
@@ -530,7 +530,7 @@ public class MusicXMLWriter{
 
 					int harmonicAdjustedValue = stringValue + noteValue + harmonicValue;
 
-					if(n > 0){
+					if (n > 0){
 						this.addNode(noteNode, "chord");
 					}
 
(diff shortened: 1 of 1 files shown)


desktop/TuxGuitar-musicxml/src/app/tuxguitar/io/musicxml/MusicXMLWriter.java
```diff
@@ -577,12 +582,10 @@ private void writeBeats(Node parent, TGMeasure measure, int nVoice, boolean meas
...
 					this.removeNodeIfNoChildren(notationsNode);
 
-					// Let-ring would be applied here... But the MusicXML 4.0 spec does not have a defined element for that.
-					// GP has it as a processing instruction, but until it is officially supported, we will leave it out.
-					// Uncomment and update when the MusicXML spec changes to support the Let-Ring status.
-					// if (note.getEffect().isLetRing()){
-					// 	this.addNode(noteNode, "letring");
-					// }
+					if (note.getEffect().isLetRing()){
+						Node tiedNode = this.addNode(noteNode, "tied");
```
guiv42, line 586: "let ring" and "tied" are different concepts. See screenshot, C is "let-ring", Es are "tied"
lr_tied

I double-checked in musicXML specification, and could not find  any equivalent for the concept of "let ring" (this is consistent with initial comment)

guiv42 (commented):
Thanks for the PR.

I just have a problem b9bc8f505ec7cff188529f5d0730d27bb03cc2a7, I think it shall be reverted.

PR #697, opened 2025-04-04
[MusicXML] export dead note

rettinghaus:
This adds export of the correct notehead for dead notes (in standard notation) and the palm mute symbol. 
Furthermore it makes the number of staves and the absence of a part symbol explicit.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -232,3 +232,7 @@ Taehun Yun
 Alejandro6626
     * Spanish translation
     * Contact: https://github.com/alejandro6626
+
+Klaus Rettinghaus
+    * MusicXML export
+    * Contact: https://github.com/rettinghaus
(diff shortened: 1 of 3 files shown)


AUTHORS
```diff
@@ -232,3 +232,7 @@ Taehun Yun
 Alejandro6626
     * Spanish translation
     * Contact: https://github.com/alejandro6626
+
+Klaus Rettinghaus
```
guiv42, line 236: You can also add your contribution in the [user doc](https://github.com/helge17/tuxguitar/blob/80545d0e5e39425f30b08f76728e485e8f7b9030/desktop/TuxGuitar/share/help/about.html#L61), in html.
This duplication of data is not very practical, but it's not that simple as the list of contributors is expected under *markdown* format by github, under *text* format by TuxGuitar (menu Help/About), and under *html* format for both [website](https://tuxguitar.app) and TuxGuitar (menu Help/Documentation)

guiv42:
If you're interested in musicXML export, just for information there is an interesting list of suggestions in #289

rettinghaus:
By default parts with more than one staff carry a brace. Here's the relevant example: <https://www.w3.org/2021/06/musicxml40/musicxml-reference/examples/staff-type-element/>

guiv42 (approved):

Issue #37, opened 2023-05-18
TuxGuitar standard E Tuning is "E3 A3 D4 G4 B4 E5" - should be "E2 A2 D3 G3 B3 E4"


helge17:
I noticed that in the tuning settings all tunings are displayed one octave too high.
While normally MIDI note 60 is equal to C4, TuxGuitar shows MIDI note 60 as C5, etc.

At least visually this can be fixed by subtracting 1 at the right place:

```
--- a/TuxGuitar/src/org/herac/tuxguitar/app/view/dialog/track/TGTrackTuningDialog.java
+++ b/TuxGuitar/src/org/herac/tuxguitar/app/view/dialog/track/TGTrackTuningDialog.java
@@ -694,9 +694,10 @@ public class TGTrackTuningDialog {
 			sb.append(NOTE_NAMES[value % NOTE_NAMES.length]);
 
 			if( octave ) {
-				sb.append(value / MAX_NOTES);
+				sb.append(value / MAX_NOTES - 1);
 			}
 		}
+
 		return sb.toString();
 	}
 
```
The MusicXML export has a similar problem as reported in https://sourceforge.net/p/tuxguitar/bugs/136/

Again, you can lower the octave by one in the following lines:

```
--- a/TuxGuitar-musicxml/src/org/herac/tuxguitar/io/musicxml/MusicXMLWriter.java
+++ b/TuxGuitar-musicxml/src/org/herac/tuxguitar/io/musicxml/MusicXMLWriter.java
@@ -194,7 +194,7 @@ public class MusicXMLWriter {
 			Node stringNode = this.addNode(staffDetailsNode, "staff-tuning");
 			this.addAttribute(stringNode, "line", Integer.toString( (track.stringCount() - string.getNumber()) + 1 ) );
 			this.addNode(stringNode, "tuning-step", NOTE_NAMES[ NOTE_SHARPS[ (string.getValue() % 12) ] ] );
-			this.addNode(stringNode, "tuning-octave", Integer.toString(string.getValue() / 12) );
+			this.addNode(stringNode, "tuning-octave", Integer.toString(string.getValue() / 12 - 1) );
 		}
 	}
 	
@@ -269,7 +269,7 @@ public class MusicXMLWriter {
 					
 					Node pitchNode = this.addNode(noteNode,"pitch");
[...]

PR #701, opened 2025-04-06
Fix clef export in MusicXML

rettinghaus:
This fixes the export of clefs:

- remove the incorrect `clef-octave-change` (adding missing `transpose` information instead)
- add `after-barline` attribute
- use correct `sign` for c clefs

Other changes just small formatting improvements.

### DIFF
--- a/desktop/TuxGuitar-musicxml/src/app/tuxguitar/io/musicxml/MusicXMLWriter.java
+++ b/desktop/TuxGuitar-musicxml/src/app/tuxguitar/io/musicxml/MusicXMLWriter.java
@@ -399,6 +399,7 @@ public class MusicXMLWriter{
 		if (!isPercussion){
 			node = this.addNode(parent, "clef");
 			this.addAttribute(node, "number", "2");
+			this.addAttribute(node, "print-object", "no");
 			this.addNode(node, "sign", "TAB");
 		}
 	}
--- a/desktop/TuxGuitar-musicxml/src/app/tuxguitar/io/musicxml/MusicXMLWriter.java
+++ b/desktop/TuxGuitar-musicxml/src/app/tuxguitar/io/musicxml/MusicXMLWriter.java
@@ -309,8 +309,11 @@ public class MusicXMLWriter{
 				this.writeClef(measureAttributes, measure.getClef(), isPercussion);
 			}
 
-			if (!isPercussion && (previous==null || measure.getNumber() == 1)){
-				this.writeTuning(measureAttributes, measure.getTrack(), measure.getKeySignature());
+			if (previous==null || measure.getNumber() == 1) {
+				if (!isPercussion){
+					this.writeTuning(measureAttributes, measure.getTrack(), measure.getKeySignature());
+				}
+				this.writeTransposition(measureAttributes);
 			}
 		}
 	}
@@ -331,6 +334,13 @@ public class MusicXMLWriter{
 		if (trackOffset > 0){
 			this.addNode(staffDetailsNode, "capo", Integer.toString( trackOffset ));
 		}
+	}
+
+	private void writeTransposition(Node parent){
+		Node transposeNode = this.addNode(parent, "transpose");
+		this.addAttribute(transposeNode, "number", "1");
+		this.addNode(transposeNode, "chromatic", "0");
+		this.addNode(transposeNode, "octave-change", "-1");
 	}
 
 	private void writeNote(Node parent, String prefix, int value, int keySignature){
@@ -360,6 +370,7 @@ public class MusicXMLWriter{
 	private void writeClef(Node parent, int clef, boolean isPercussion){
 		// first clef: score
 		Node node = this.addNode(parent, "clef");
+		this.addAttribute(node, "after-barline", "yes");
 		if (!isPercussion){
 			this.addAttribute(node, "number", "1");
 		}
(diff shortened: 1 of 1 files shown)


desktop/TuxGuitar-musicxml/src/app/tuxguitar/io/musicxml/MusicXMLWriter.java
```diff
@@ -370,20 +381,18 @@ private void writeClef(Node parent, int clef, boolean isPercussion){
 		else if(clef == TGMeasure.CLEF_TREBLE){
 			this.addNode(node, "sign", "G");
 			this.addNode(node, "line", "2");
-			this.addNode(node, "clef-octave-change", String.valueOf(-1));
```
guiv42, line 373: I think the original implementation is correct.
TuxGuitar is dedicated to guitar, and guitar *is* a transposing instrument, using a key 1 octave below the standard one (same for bass guitar).
Where TuxGuitar is wrong: it uses this convention silently, and doesn't display an explicit "8" below the clefs (as musescore does: just create an empty guitar track, the "8" is correctly shown)
If I'm not wrong, Musescore 4.5.1 uses the same "clef-octave-change" musicxml node.

See screenshots.
left : TuxGuitar
right: exported by TG1.6.6 in musicxml, and imported by musescore 4.5.1 (correct)

f_tg  f_tg166_mu

After applying your PR, and exporting to musicxml, this is  how it appears in musescore (incorrect):

f_mu

desktop/TuxGuitar-musicxml/src/app/tuxguitar/io/musicxml/MusicXMLWriter.java
```diff
@@ -399,6 +399,7 @@ else if(clef == TGMeasure.CLEF_ALTO){
 		if (!isPercussion){
 			node = this.addNode(parent, "clef");
 			this.addAttribute(node, "number", "2");
+			this.addAttribute(node, "print-object", "no");
```
guiv42, line 402: why?
see screenshots above, original implementation seems correct to me

desktop/TuxGuitar-musicxml/src/app/tuxguitar/io/musicxml/MusicXMLWriter.java
rettinghaus: We cannot use ottava clefs in export, because a) these are not the clefs TuxGuitar is using, and b) that wouldn't work with the c clefs (wich are not exported correctly currently).

grafik

Correct, guitar is a transposing instrument, but that wasn't shown in the export. The shown pitch is for both clefs the same, even if the old implementation showed the note on the "correct line". 
The transposition information in this PR is now correct, but I need to adjust the pitch for the score staff.

desktop/TuxGuitar-musicxml/src/app/tuxguitar/io/musicxml/MusicXMLWriter.java
rettinghaus: TuxGuitar doesn't show a tab clef, so it makes sense to make it expliciti in the export that this _is_ a tab staff, but do not show the clef.

desktop/TuxGuitar-musicxml/src/app/tuxguitar/io/musicxml/MusicXMLWriter.java
guiv42: OK. Thanks for your contributions, I must admit I don't have a good knowledge of musicXML data structure.

desktop/TuxGuitar-musicxml/src/app/tuxguitar/io/musicxml/MusicXMLWriter.java
guiv42: I think it's better if we keep it explicit in the musicXML exported files. In fact, the cleanest option would probably be to make it explicit in TuxGuitar also. Unfortunately it's not the simplest option, as I don't really understand how the clefs are currently [painted](https://github.com/helge17/tuxguitar/blob/bef6d00a8efdafdf2355fc5b5ebf5eecd5fae241/common/TuxGuitar-lib/src/main/java/app/tuxguitar/graphics/control/painters/TGClefPainter.java#L13). More precisely I don't understand how to create such a code.

guiv42:
I'm sorry but I think I still need some explanations.
See the screenshot below, done after your last commit (317973b). It's the same test as above, just a F on 1sr fret, 1st string. Exported by TuxGuitar in musicXML, then imported by MuseScore 4.5.1:


The note appears on the correct line, however the representation of the clef omits the explicit "8" below the clef (it appears correctly before the PR). Why remove the "clef-octave-change" musicxml node?

guiv42:
It looks like a conflict of octave numbering convention. TuxGuitar uses midi convention, so A**4**=**440**Hz.
And 440Hz = 1st string 5th fret in standard tuning. See #37. It looks like Musescore uses the same convention for musicXML

rettinghaus:
The MuseScore import is not correct, it cannot handle different transpositions in one part (yet). 

A violin clef marks the G4 on the second line, the 8 below the clef changes the written octave, so then it marks the G3. <https://www.dacapoalcoda.com/octaves-clefs>. The guitar is a transposing instrument sounding one ocatve lower than written, i.e. this would the sound like G2. MIDI does not know about about transposing instruments, but refelcts the sounding pitch.

guiv42:
Let me take one step back (I'm trying to understand).
Say I want to play one **A 440Hz** on a standard tuned guitar. What should it look like in musicXML, I don't know!
So, I tried  an empiric approach: I exported this test song with TuxGuitar 1.6.6 and latest commit of this PR, then I opened the produced musicxml files with available viewers:

| Viewer | TuxGuitar 1.6.6 | commit 181530b |
| ----------- | ----------- | ----------- |
| MuseScore 4.5.1 | musescore451  | musescore451  |
| [soundslice.com](https://www.soundslice.com/musicxml-viewer/) | soundslice |  soundslice |
| [opensheetmusicdisplay.org](https://opensheetmusicdisplay.org/public-ts-demo/) | opensheetmusicdisplay  | opensheetmusicdisplay |

My conclusions:
- I'm still confused!
- the only 100% correct configuration is TG1.6.6 + MuseScore 4.5.1
- on the 2 web viewers it clearly looks better after commit 181530b, only the "8" is missing


@rettinghaus, two questions: 
1. if MuseScore is incorrect, could you eventually create an issue in MuseScore's repo?
2. why not add the `<clef-octave-change>` musicxml node?

rettinghaus:
As said before, `clef-octave-change` changes the octave of the written notes, not their resulting sounding pitch. It's something different than transposition.

But this needs some more work. I'll open an issue on MuseScore side.

guiv42:
Sorry but I still don't understand. Naively I would assume that the pitch is independent from the content of `<clef>` node, and only defined by:
```
                <pitch>
                    <step>A</step>
                    <octave>5</octave>
                </pitch>
```

rettinghaus:
The `pitch` node gives the _notated_ pitch, the sounding pitch is dependent on the information given in the `transposition` node (if existend).

The display of the written pitch is then calculated for the given clef, which depends on it's form, the line it sits on and the given `clef-octave-change`.

PR #705, opened 2025-04-11
Update Polish translation

Pruchniakowski:
Update Polish translation
Many fixes in this file.

Greetings
Grzegorz


guiv42:
All files and folders in "deprecated" folder are, hmm... deprecated!
I moved them there initially because I was not 100% sure of what these modules were. It seems it's been counter-productive: you're not the first one trying to patch something there, but it's useless.
I'll just delete the whole folder (which remains anyway available in git history)

PR #706, opened 2025-04-11
cleanup, delete deprecated items

guiv42:
to avoid attempts from contributors to patch these unused items
see #705, #690

@helge17 : is it OK for you?


helge17 (approved):
It's OK for me to delete this.

PR #711, opened 2025-04-13
free edition mode

guiv42:
see #579


guiv42:
Last regressions were fixed a couple of weeks ago, no new issue detected.
I think we can now merge tuxguitar-next branch (and even delete it after?)
@helge17 , what do you think?

helge17 (approved):
That's fine! I will remove tuxguitar-next in a few days.

Issue #698, opened 2025-04-05
make isAnyTiedTo easily available


rettinghaus:
In LilyPond and ABC output there is a function to check if the next note is tied to the current one. We need this for MusicXML export too (and maybe in some more places, see #485). 
Before shamelessly copying it over to the MusicXML exporter maybe it would be good to make this function more globally available, maybe even make it a method in the note class itself?
What do you think?

guiv42:
Your question makes perfectly sense, I already had similar issues.

As far as I understand the code base, `TGNote` class and all [models](https://github.com/helge17/tuxguitar/tree/master/common/TuxGuitar-lib/src/main/java/app/tuxguitar/song/models) are only *flat* data structures: `TGNote` methods don't use other model classes methods or attributes (like `TGEffect`). There are links between models in child class [`TGNoteImpl`](https://github.com/helge17/tuxguitar/blob/master/common/TuxGuitar-lib/src/main/java/app/tuxguitar/graphics/control/TGMeasureImpl.java) (e.g. a [getNoteForTie()](https://github.com/helge17/tuxguitar/blob/80545d0e5e39425f30b08f76728e485e8f7b9030/common/TuxGuitar-lib/src/main/java/app/tuxguitar/graphics/control/TGNoteImpl.java#L584) method, similar to the one you're looking for), but often we get an instance of the parent class `TGNote` to work with. This leads to (questionable) casts in many places, that occasionally even [fail](https://github.com/helge17/tuxguitar/discussions/380). Methods requiring links between models are generally implemented in a [manager](https://github.com/helge17/tuxguitar/tree/master/common/TuxGuitar-lib/src/main/java/app/tuxguitar/song/managers)  class.
This is my understanding, and I've never been comfortable with this (historical) data structure. No criticism here, I'm not a software architect.

Proposal: create the method you need in `TGTrackManager`, which could then be instantiated in musicXML exporter. 
Already existing methods should help (e.g. `TGMeasureManager.getNextVoice()`). I can do it if you want.
Any other suggestion welcome.

rettinghaus:
Sounds great, please go ahead. I don't know about the structure too much here, but I'm happy to assist.

guiv42:
It's not as obvious at it looks. Since "tied" is only a boolean attribute attached to a note, it's currently easy to define *invalid* ties.
See screenshot:

![Image](https://github.com/user-attachments/assets/e195f878-e153-42eb-bbdb-5f71d7869cfa)

[...]

PR #712, opened 2025-04-15
ability to retrieve next tied note

guiv42:
yet unused
need identified for musicXML exporter, but could be used more widely

see #698


rettinghaus (approved):
Looks good to me, thank you!

PR #719, opened 2025-04-20
Improve doc with corrected path

G-eos:
Documentation path seems not accurate. Slight proposed fix to be coherent with the repository file structure.


guiv42:
Not understood.
From my web browser, if I display the following page https://github.com/helge17/tuxguitar/blob/master/doc/contribute.md then click on one link to "install.md" this works fine.
(note: there are symlinks)

G-eos:
Yes ! Thank you.
It works well for me too without this un-needed change. I note, that symlinks does not appears easily on GitHub web neither in VisualStudioCode.
I will close the PR as useless.

PR #720, opened 2025-04-20
Update MacOS aarch64 (Mac Mx) build instruction

G-eos:
Install instruction modification to download correct file for SWT.

On my M3 MacBook, following INSTALL.md with the proposed modification, I succesfully build TuxGuitar. Using openjdk from homebrew it starts well and run perfectly :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ target/
 .project
 android/build-scripts/tuxguitar-android/local.properties
 android/build-scripts/tuxguitar-android/.idea/
+.DS_Store
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -138,10 +138,11 @@ $ brew install openjdk maven wget
 ### Download and install SWT for macOS
 
 ```sh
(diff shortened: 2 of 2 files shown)


guiv42 (commented):
just 1 question: why modify .gitignore?

G-eos:
To filter out standard MacOS auto generated files to have a cleaner fit status (it adds a file in every opened directory :-( )
See: https://en.wikipedia.org/wiki/.DS_Store?wprov=sfti1

As these PR is MacOS related I add it. I clearly understand that it has low meaning, tell me if you want I remove it.

guiv42:
That's fine for me.
@helge17: if you don't mind I will let you merge this one, as I cannot test. Possibly your build script would also need to be modified for consistency?

guiv42 (approved):

helge17:
The changes seem reasonable. Currently I can't build on an ARM Mac so I'm not adapting the build script yet.

PR #741, opened 2025-04-27
trying to handle invalid tied notes

guiv42:
@helge17, I would like your opinion on this PR (on the feature itself, not necessarily on the code)

The idea is to try to handle *invalid tied notes* quite smoothly.
Theoretically, 2 tied notes should have same {pitch, string, fret} and be immediately consecutive.
However, I found many invalid tied notes in different tabs, so I assume they are sometimes used to represent more easily something that should be written with several voices.

So, the idea is to limit the creation of invalid tied notes, but still to enable it if this is intentional. I re-used the *free edition mode*, and the *measure errors dialog* for this purpose.

Notes:
- I had to revert one of your recent commits (translation of "fix voice", that is now "fix error")
- user doc is not yet updated
- android version is not yet updated

demo:

https://github.com/user-attachments/assets/a4677d5e-0a9b-4f4d-b408-f413536b807d


helge17 (approved):
As my son is the real guitar hero in the family, I asked him: he finds the feature logical and useful.

When the free edition mode is finished, I will rework the docs.

guiv42:
OK, thank you.

Do you have any idea why building the Android version fails?
In my config, it fails when trying to build [TestTiedNotes.java](https://github.com/guiv42/tuxguitar/blob/invalidTiedNotes/common/TuxGuitar-editor-utils/src/test/java/app/tuxguitar/editor/TestTiedNotes.java), because of a missing dependency to junit. If I delete manually this file (or comment out its content) then Android app builds successfully.
I don't understand why AndroidStudio tries to build that file, which is in a `src/test/` folder. It's there only as a unit test, to be executed by maven when building desktop version. I don't have the same issue with test classes for[ tuxguitar-lib](https://github.com/guiv42/tuxguitar/tree/invalidTiedNotes/common/TuxGuitar-lib/src/test/java/app/tuxguitar).

If I cannot solve this, I might as well delete this test, and restore the original folders structure.

PR #746, opened 2025-05-01
Create a GitHub Action to build  for Android

G-eos:
Please find an Android CI to build TuxGuitar. Merge this PR if its helps, I do not use Android myself so I am unable to test.

Java version 21 is selected but it may not be the accurate one.

### DIFF
--- a/.github/workflows/android-gradle.yml
+++ b/.github/workflows/android-gradle.yml
@@ -31,7 +31,7 @@ jobs:
     - uses: actions/upload-artifact@v4
       with:
         name: Package-Android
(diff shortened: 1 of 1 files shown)


guiv42:
I like the idea: I already forgot several times to update Android app following modification of desktop app, that would have been detected with this PR.
I just tested, it produces an apk file, but I cannot install it on my Android device: "package seems to be invalid" (translated message).

However, I consider that the objective of such a workflow is essentially to provide a test for compatibility with Android build. Not to produce valid apk files.

So I'm favorable to merging this PR. @helge17 what is your opinion?

PR #750, opened 2025-05-04
unify casts

rettinghaus:
A small PR just to clean up and unify the style of the casts happening, removing unneeded parentheses and imrpoving readability.


guiv42:
I have a problem with this PR: there are too many modified files, and I don't know how to check this. There are no tests we can rely on to catch an eventual error

@rettinghaus: if you want to improve the formatting, I suggest to do it on a much smaller scope. E.g. apply the [formatting rules](https://github.com/helge17/tuxguitar/blob/master/misc/tuxguitar_formatting.xml) to a set a files when implementing a new feature, so that these modifications are covered by the tests

rettinghaus:
Ok, closing this one to break it down into smaller chunks.

PR #751, opened 2025-05-04
add macos 15 to ci

rettinghaus:


G-eos:
I was evaluating the same in parallel in https://github.com/G-eos/tuxguitar/tree/refs/heads/workflow-mac15
After download and comparaison of the generated build the result is:
- macOS-14 and macOS-15 generate the same files (after opening .jar files for comparison of contents)
- macOS-13 generated files are the same except for swt.jar (which come from external dependancies pulled during the build process). The differences comes from the macOS-13 which is an Intel x86 platform which pull swt-x86 versus macOS-14/15 which pull an swt for aarch64/arm64.

My advice would be to keep macOS-13 for x86 legacy and macOS-15 for aarch64

Issue #760, opened 2025-05-06
Left-to-right selection breaks in invalid measure


helge17:
When selecting some measures from left to right in free edition mode, the selection breaks at the end of the invalid measure. Also it is not possible to navigate to the surplus beat with the arrow keys or the mouse.

The selection from right to left works.

https://github.com/user-attachments/assets/a4c4d961-ec51-41bb-83b2-f81c52e5cc29

guiv42:
First of all: thank you @helge17 for all the tests and the reported issues.

Discontinuity of selection was introduced deliberately to fix a crash scenario (#733). Selection explicitly assumes that selected beats are chronologically ordered, and that cannot always be true when measures are too long: in the screenshots above, last beat of measure 2 starts after the first beat of measure 3.

In fact it was unexpected that it worked right-to-left, and #761"solved" that: selection now breaks in both directions.
I don't see any other solution than documenting this behavior: selection can't be extended beyond an invalid measure.

In fact it's pretty difficult to use the free edition mode with multiple voices: it creates very easily overlaps that are difficult to understand. In the screenshots, B in measure 2 starts when C# ends, and that's right in the middle of the A#. This overlap is not trivial to see / to fix.
I'm currently investigating an alternative solution.

PR #763, opened 2025-05-06
Free edition mode, simplify multi-voices

guiv42:
@helge17: some testing would be appreciated if you can. I hope this should simplify the management of multi-voices in free edition mode.

Basically, the idea is to prefer "holes" in a voice than overlaps. These are easier to fix, as TG tries to fill them automatically with rests, and succeeds most of the times.
With this, your test file used in #760 should look very different. Here is an illustration of the difference before/after this PR.
In this test, one note duration is increased, first in normal mode, then in freeEditionMode. Before the PR, this creates an overlap with following beat (00:09 in video). After the PR, this inserts a rest in the other voice (00:25). That should be more understandable, and easier to fix.

https://github.com/user-attachments/assets/9d320ee8-c93e-4c9f-af53-78060367a09d

This PR also includes another modification: it's now easier to fix too long measures ending with rests (just deleting the last rest now works also in multi-voices)


helge17:
I think the behaviour with this PR is better than before.

Before:
grafik

With PR:
grafik

It is more logical for voice 1 to wait until the full note in voice 2 is finished before continuing, so that both voices are in sync again afterwards.

Also, with the PR it is easier to select the surplus beats and navigate over them with the arrow keys.

I would suggest merging this PR. Then I can test the Free edition mode again in more detail.

guiv42:
thanks!

PR #764, opened 2025-05-06
android workflow: remove apk

guiv42:
apk is useless, as not signed
without this, the test is sufficient to check Android app builds correctly

### DIFF
--- a/.github/workflows/android-gradle.yml
+++ b/.github/workflows/android-gradle.yml
@@ -28,10 +28,6 @@ jobs:
         cache: gradle
     - name: Build the app
       run: cd android/build-scripts/tuxguitar-android && ./gradlew && ./gradlew assembleRelease
-    - uses: actions/upload-artifact@v4
-      with:
-        name: Package-Android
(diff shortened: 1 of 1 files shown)


helge17:
If it is only the missing signature that makes the apk uninstallable, the apk can be signed with a self-signed key. See [this](https://github.com/helge17/tuxguitar/blob/e9d28a0e9479248df570514db0e3e71ca90bd172/misc/build_tuxguitar_from_source.sh#L530-L533) comment in the build script on how to create the key and a few lines [below](https://github.com/helge17/tuxguitar/blob/e9d28a0e9479248df570514db0e3e71ca90bd172/misc/build_tuxguitar_from_source.sh#L545) how to sign the apk.

guiv42:
I don't think it's worth the effort.
Absence of signature is for sure a problem, but it may not be the only one. If this apk is signed with another key than the official delivery, users will have problems when trying to update from one to another. And I don't want to spend time on user support for that.
The most important is OK: if someone pushes a modification that breaks Android build (as I already did in the past), then it will be detected

PR #783, opened 2025-05-28
Add option to set how long the count-in is in Player 

tdan:
- Add ability to set and get count-in tick in `MidiPlayerCountDown`
- Add count-in tick option to `TGTransportDialog`

### DIFF
--- a/desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/transport/TGTransportDialog.java
+++ b/desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/transport/TGTransportDialog.java
@@ -13,6 +13,7 @@ import app.tuxguitar.app.action.impl.measure.TGGoLastMeasureAction;
 import app.tuxguitar.app.action.impl.measure.TGGoNextMeasureAction;
 import app.tuxguitar.app.action.impl.measure.TGGoPreviousMeasureAction;
 import app.tuxguitar.app.action.impl.transport.TGOpenTransportModeDialogAction;
+import app.tuxguitar.app.action.impl.transport.TGTransportCountDownAction;
 import app.tuxguitar.app.action.impl.transport.TGTransportMetronomeAction;
 import app.tuxguitar.app.action.impl.transport.TGTransportPlayPauseAction;
 import app.tuxguitar.app.action.impl.transport.TGTransportStopAction;
@@ -47,6 +48,8 @@ import app.tuxguitar.ui.event.UIMouseDownListener;
 import app.tuxguitar.ui.event.UIMouseEvent;
 import app.tuxguitar.ui.event.UIMouseMoveListener;
 import app.tuxguitar.ui.event.UIMouseUpListener;
+import app.tuxguitar.ui.event.UISelectionEvent;
+import app.tuxguitar.ui.event.UISelectionListener;
 import app.tuxguitar.ui.layout.UITableLayout;
 import app.tuxguitar.ui.resource.UIColor;
 import app.tuxguitar.ui.resource.UIColorModel;
@@ -59,6 +62,7 @@ import app.tuxguitar.ui.widget.UILabel;
 import app.tuxguitar.ui.widget.UILayoutContainer;
 import app.tuxguitar.ui.widget.UIPanel;
 import app.tuxguitar.ui.widget.UIProgressBar;
+import app.tuxguitar.ui.widget.UISpinner;
 import app.tuxguitar.ui.widget.UIToggleButton;
 import app.tuxguitar.ui.widget.UIWindow;
 import app.tuxguitar.util.TGContext;
@@ -82,6 +86,8 @@ public class TGTransportDialog implements TGEventListener {
 	private UILabel label;
 	private UIProgressBar tickProgress;
 	private UIToggleButton metronome;
+	private UIToggleButton countInToggle;
+	private UISpinner countInTicks;
 	private UIButton mode;
 	private UIToolBar toolBar;
 	private UIToolActionItem first;
@@ -167,6 +173,36 @@ public class TGTransportDialog implements TGEventListener {
 		this.mode = factory.createButton(composite);
 		this.mode.addSelectionListener(new TGActionProcessorListener(this.context , TGOpenTransportModeDialogAction.NAME));
 		compositeLayout.set(this.mode, 2, 1, UITableLayout.ALIGN_FILL, UITableLayout.ALIGN_FILL, false, true);
+		
+		MidiPlayer player = MidiPlayer.getInstance(this.context);
+		
+		this.countInToggle = factory.createToggleButton(composite);
+		this.countInToggle.setToolTipText(TuxGuitar.getProperty("transport.count-down"));
+		this.countInToggle.setSelected(player.getCountDown().isEnabled());
+		this.countInToggle.addSelectionListener(new TGActionProcessorListener(this.context, TGTransportCountDownAction.NAME));
+		
+		this.countInTicks = factory.createSpinner(composite);
+		this.countInTicks.setEnabled(player.getCountDown().isEnabled());
+		if (player.getCountDown().getTickCount() == 0)
+			this.countInTicks.setValue(player.getSong().getMeasureHeader(0).getTimeSignature().getNumerator());
+		else
+			this.countInTicks.setValue(player.getCountDown().getTickCount());
+		
+		this.countInTicks.addSelectionListener(new UISelectionListener() {
+			@Override
+			public void onSelect(UISelectionEvent event) {
+				player.getCountDown().setTickCount(countInTicks.getValue());
+			}
+		});
+		this.countInToggle.addSelectionListener(new UISelectionListener() {
+			@Override
+			public void onSelect(UISelectionEvent event) {
+				countInTicks.setEnabled(!player.getCountDown().isEnabled());
+			}
+		});
+		
+		compositeLayout.set(this.countInToggle, 1, 2, UITableLayout.ALIGN_FILL, UITableLayout.ALIGN_FILL, false, true);
+		compositeLayout.set(countInTicks, 2, 2, UITableLayout.ALIGN_RIGHT, UITableLayout.ALIGN_CENTER, false, false);
 
 		this.loadOptionIcons();
 	}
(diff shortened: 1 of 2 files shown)


desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/transport/TGTransportDialog.java
```diff
@@ -167,6 +173,36 @@ private void initOptions(UILayoutContainer parent) {
...
+		
+		this.countInToggle = factory.createToggleButton(composite);
+		this.countInToggle.setToolTipText(TuxGuitar.getProperty("transport.count-down"));
+		this.countInToggle.setSelected(player.getCountDown().isEnabled());
+		this.countInToggle.addSelectionListener(new TGActionProcessorListener(this.context, TGTransportCountDownAction.NAME));
+		
+		this.countInTicks = factory.createSpinner(composite);
+		this.countInTicks.setEnabled(player.getCountDown().isEnabled());
+		if (player.getCountDown().getTickCount() == 0)
+			this.countInTicks.setValue(player.getSong().getMeasureHeader(0).getTimeSignature().getNumerator());
```
guiv42, line 187: why the first measure?

desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/transport/TGTransportDialog.java
```diff
@@ -167,6 +173,36 @@ private void initOptions(UILayoutContainer parent) {
...
+		else
+			this.countInTicks.setValue(player.getCountDown().getTickCount());
+		
+		this.countInTicks.addSelectionListener(new UISelectionListener() {
+			@Override
+			public void onSelect(UISelectionEvent event) {
+				player.getCountDown().setTickCount(countInTicks.getValue());
```
guiv42, line 194: It's not obvious, I've got to think about it. That would be probably better if done asynchronously (i.e. through an actionProcessor, see an example in line 182).
With this design, if someone wants to add this spinner elsewhere in the ui later (toolbar/other dialog) it won't be notified by any event when value changes and cannot be updated.

desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/transport/TGTransportDialog.java
```diff
@@ -167,6 +173,36 @@ private void initOptions(UILayoutContainer parent) {
...
+			}
+		});
+		this.countInToggle.addSelectionListener(new UISelectionListener() {
```
guiv42, line 197: I don't understand, you have already created a selectionListener at line 182
And: same problem as above, this is not the right place to update this attribute

guiv42 (changes requested):
Thanks for the contribution, I'd say it's a good starting point.

Take care to understand how the different UI elements work.
General hint: whenever you work in an area of code you don't master, it's highly preferable to mimic an already existing implementation. E.g. in this case, the count-in toggle button has a *very similar* behavior as the metronome button. That should have led you to update the associated attributes (toolTipText, enabled, etc) the same way as for the metronome.

This dialog can remain open permanently, so its content shall be dynamically updated (not only set when it is displayed).
Like many menus/dialogs/toolbars, this dialog is listening to events fired by TG event manager (see line 125). Dialog content shall be updated when such events are received.
As an example: you did it correctly for the icon of count-in button (set at the same place as the metronome). Therefore, it's updated dynamically when user changes skin with this dialog already opened.

Don't hesitate to ask if you need some support.

desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/transport/TGTransportDialog.java
```diff
@@ -167,6 +173,36 @@ private void initOptions(UILayoutContainer parent) {
...
 		compositeLayout.set(this.mode, 2, 1, UITableLayout.ALIGN_FILL, UITableLayout.ALIGN_FILL, false, true);
+		
+		MidiPlayer player = MidiPlayer.getInstance(this.context);
+		
+		this.countInToggle = factory.createToggleButton(composite);
+		this.countInToggle.setToolTipText(TuxGuitar.getProperty("transport.count-down"));
+		this.countInToggle.setSelected(player.getCountDown().isEnabled());
+		this.countInToggle.addSelectionListener(new TGActionProcessorListener(this.context, TGTransportCountDownAction.NAME));
+		
+		this.countInTicks = factory.createSpinner(composite);
```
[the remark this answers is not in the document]
tdan: - Added Min value 1
- [UISpinner](https://github.com/helge17/tuxguitar/blob/master/desktop/TuxGuitar-ui-toolkit/src/app/tuxguitar/ui/widget/UISpinner.java) doesn't have setToolTip() method. Would you like me to add this functionality as well? (maybe in a separate PR since we will need to add concrete implementation of it)

desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/transport/TGTransportDialog.java
tdan: I had to create this listener to enable spinner every time toggle button was selected.

This isn't necessary anymore since updating attributes are put in the right place. Everything seems to work seamlessly

desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/transport/TGTransportDialog.java
tdan: I've tried other ways to get current measure time. However, if I don't play a loop at least once, it will give me a NullPointerException. I found first measure to be the most reliable way to get a default count-in value. Even the start-up empty sheet has its first measure anyway.

We can leave it as default  0. I think it's just weird to have count-in plays a # of counts but spinner is set to 0.

I think having count-in ticks update asynchronously will solve this issue. I'm working on it.

desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/transport/TGTransportDialog.java
tdan: Agree 100% this is the way to go.

desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/transport/TGTransportDialog.java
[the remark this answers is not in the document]
guiv42: you're right about the toolTipText, just forget it

desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/transport/TGTransportDialog.java
tdan: I've made a few changes and able to keep value adapted to caret position, that is before count-in is enabled for the first time. After that, the value is retained from previous behavior regardless of caret position. 

For example:

- Measure 1 is 3/4, measure 10 is 8/4
- Move caret to measure 10, value's adapted o 8
- Enable count-in and play the measure for the first time. 8 beats of count down as desired.
- Move back to measure 1, disable & enable count-in again. Value stays at 8 (in fact it will remain like that till user actively changes the value)

Question: is it the desired behavior? Or do we need to reset tick count behavior every time count-in is disabled? 

I can see usage of both ways.

guiv42:
@tdan : I see you are currently working on this PR. Please request explicitly a review when you are ready (I won't have a look before that)

desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/transport/TGTransportDialog.java
guiv42: > Each time the spinner value changes, it updates the ticks value, and any dialog that opens will read the most recent updated value

Yew, but this is only true when a new dialog is opened. If we can have the same indicator simultaneously visible in 2 places, acting on one will not update the second. E.g. if we want to add it in a toolbar one day, the toolbar will not be notified when count-in value changes in the dialog. As an illustration, you can see the volume controls available in tracks table (bottom left) and in channels dialog (View/Show instruments): whenever a value changes in one place, the second is updated also.
Anyway, this is a rather complex evolution (see #176) that may not be needed (at least in a short term). OK to keep current implementation.

desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/transport/TGTransportDialog.java
```diff
@@ -316,6 +338,19 @@ public void updateItems(boolean force){
 			this.tickProgress.setMinimum((int)first.getStart());
 			this.tickProgress.setMaximum((int)(last.getStart() + last.getLength()) -1);
 			this.metronome.setSelected(player.isMetronomeEnabled());
+			this.countInToggle.setSelected(player.getCountDown().isEnabled());
+			this.countInTicks.setMinimum(1);
```
guiv42, line 342: Min value is a constant attribute of `countInTicks`, no need to update the value: it can be set once for all when spinner is created.
Move `.setMinimum(1)` to `initOptions()`, just after creation of `countInTicks`

desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/transport/TGTransportDialog.java
```diff
@@ -316,6 +338,19 @@ public void updateItems(boolean force){
...
+			// Set default spinner's value to first measure time signature or current count-in ticks.
+			// Default value will automatically update to correct count when user changes song,
+			// or play a section with different time signature.
+			
+			TGMeasureHeader currentHeader = getSongManager().getMeasureHeaderAt(player.getSong(), TablatureEditor.getInstance(this.context).getTablature().getCaret().getPosition());
```
guiv42, line 347: `currentHeader` is only used in the case where `player.getCountDown().getTickCount() == 0`
So, move this into the following `if` statement (and please add explicit `{}` in your `if / else` statements, for readability)

guiv42 (commented):
just 2 minor remarks and it should be OK

tdan:
> just 2 minor remarks and it should be OK

@guiv42 just pushed these minor changes. Please take a look and let me know if there's anything else.

guiv42:
That's OK
please next time try to avoid mixing spaces and tabs for indentation (see your last commit). I'll fix it

PR #799, opened 2025-06-01
[macos][audiounit] Allow customization of the soundbank

jdemeule:
AudioUnit support SF2 sound bank, this commit allow customization of it with tuxguitar-synth-audiounit.soundbank.path variable from tuxguitar-synth-audiounit.cfg configuration file.

In addition, small "bugfix" to call send NoteOff MIDI event during noteOff call.

### DIFF
--- a/desktop/TuxGuitar-AudioUnit/jni/app_tuxguitar_player_impl_midiport_audiounit_MidiReceiverJNI.cpp
+++ b/desktop/TuxGuitar-AudioUnit/jni/app_tuxguitar_player_impl_midiport_audiounit_MidiReceiverJNI.cpp
@@ -79,26 +79,10 @@ class MidiPlayer
 	public:
 
 	// ----------------------------------------------------------------------------------------
-	MidiPlayer(char* bankPath = 0)
+	MidiPlayer(const char* bankPath = 0)
 	{
 		graph = 0;
 		__Require_noErr (result = this->createAUGraph (graph, synthUnit), ctor_home);
-
-		// if the user supplies a sound bank, we'll set that before we initialize and start playing
-		if (bankPath)
-		{
-		CFURLRef url = CFURLCreateFromFileSystemRepresentation(
-													kCFAllocatorDefault,
-													(const UInt8*)bankPath,
-													strlen(bankPath), false);
-
-		printf ("Setting Sound Bank:%s\n", bankPath);
-
-		__Require_noErr (result = AudioUnitSetProperty(synthUnit,
-									kMusicDeviceProperty_SoundBankURL,
-									kAudioUnitScope_Global, 0,
-									&url, sizeof(url)), ctor_home);
-		}
 
 		// initialize and start the graph
 		__Require_noErr (result = AUGraphInitialize (graph), ctor_home);
@@ -130,6 +114,20 @@ class MidiPlayer
 		}
 	}
 
+	OSStatus changeSoundBank(const char *bankPath) {
+		CFURLRef url = CFURLCreateFromFileSystemRepresentation(
+										kCFAllocatorDefault,
+										(const UInt8*)bankPath,
+										strlen(bankPath), false);
+
+		// printf ("Setting Sound Bank:%s\n", bankPath);
+
+		OSStatus result = AudioUnitSetProperty(synthUnit, kMusicDeviceProperty_SoundBankURL,
+															kAudioUnitScope_Global, 0, &url, sizeof(url));
+		CFRelease(url);
+      return result;
+	}
+
 	// ----------------------------------------------------------------------------------------
 	/** Note on event */
 	void noteOn(UInt32 noteNum, UInt32 onVelocity, UInt8 midiChannelInUse)
@@ -147,7 +145,7 @@ class MidiPlayer
 	/** Note off event */
 	void noteOff(UInt32 noteNum, UInt8 midiChannelInUse)
 	{
-		UInt32 noteOffCommand = kMidiMessage_NoteOn << 4 | midiChannelInUse;
+		UInt32 noteOffCommand = kMidiMessage_NoteOff << 4 | midiChannelInUse;
 		__Require_noErr (result = MusicDeviceMIDIEvent(synthUnit, noteOffCommand, noteNum, 0, 0), home_note_off);
 
 		return;
@@ -306,6 +304,14 @@ JNIEXPORT void JNICALL Java_app_tuxguitar_player_impl_midiport_audiounit_MidiRec
 JNIEXPORT void JNICALL Java_app_tuxguitar_player_impl_midiport_audiounit_MidiReceiverJNI_closeDevice(JNIEnv* env, jobject obj)
 {
 	free();
+}
+
+JNIEXPORT jint JNICALL
+Java_app_tuxguitar_player_impl_midiport_audiounit_MidiReceiverJNI_changeSoundBank(JNIEnv* env, jobject obj, jstring soundbankPath) {
+	const char* bankPath = env->GetStringUTFChars(soundbankPath, 0);
+	jint status = player->changeSoundBank(bankPath);
+	env->ReleaseStringUTFChars(soundbankPath, bankPath);
+	return status;
 }
 
 JNIEXPORT void JNICALL Java_app_tuxguitar_player_impl_midiport_audiounit_MidiReceiverJNI_noteOn(JNIEnv* env, jobject ojb, jint channel, jint note, jint velocity)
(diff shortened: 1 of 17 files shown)


guiv42:
It's for macOS, so I just can't do anything. @helge17, could you eventually have a look?

I understand this PR wants to provide the ability to configure the soundbank by manual edition of a config file.
At least this should be described in some user documentation, or the feature will be completely invisible to users. Or even better, wouldn't it be preferable to access this parameter with a settings dialog? See other plugins: some of them can be configured with "Configure" button from plugins dialog

jdemeule:
Indeed, this introduce a way to customise the soundbank used in AudioUnit midi plugin and yes, it is preferable to have it configurable with some UI.

That lead me to 2 questions:

- as the configuration is 99% identical to the jsa plugin (settings/MidiSettingsPlugin.java, settings/MidiSettingsHandler.java, utils/MidiConfigUtils.java), and probably other midi renderer could also benefit from it, should the configuration be moved at the level of the app (i.e. desktop/TuxGuitar), or do you prefer to copy paste this code?
- configuration directly open to user is subject to mistake. For now, the code is designed to failed, print a log in STDERR and does not fallback to the default soundbank. Is that something you prefer to keep, or do you think of something else?

guiv42:
> should the configuration be moved at the level of the app

It depends what you mean with that. I think it makes sense to keep each plugin independent (including its configuration). Now, if some code can be shared, possibly it could make sense to define some classes at app level, and call them only from the plugins? I don't really know, I'd say do what you think is best.

> For now, the code is designed to failed, print a log in STDERR and does not fallback to the default soundbank

I don't know exactly what possible mistakes you refer to. What I can think of:
- user defines a link to a soundbank file, and later deletes that file
- user defines a link to an invalid soundbank file (or a file which isn't a soundbank)

In both cases I think it could be useful either to add some resilience (fallback to default), or to provide an explicit and understandable error message in a dialog (or both!). Just printing to stderr might not be clear enough for most of users

jdemeule:
Clear, I will try to update the code with that direction.

jdemeule:
PR updated as suggested.

guiv42 (commented):
I'm not comfortable to review this PR, for several reasons:

1. I'm not sure it's ready for review (see my comment about unused messages)
2. It's for macos, a configuration that I cannot test. I just checked with github actions that it builds correctly for macOS 13 and 15, it's OK
3. I must admit I never opened any part of code related to sound production, so I don't feel skilled enough to perform a serious review

That being said, I see nothing problematic in this PR. Just one remark: it could be useful to update user doc (platform-dependent_plugins section of tools_sound.html?)

@jdemeule : can you please confirm the status of this PR (ready?)

helge17:
Commit 118841dd1728ec5dfcc3c8faed40b09c5b9c71be adds warning windows to the JSA plugin in case the user enters a non-existent soundbank file or if the soundbank cannot be loaded on startup. It's not perfect, but it should reduce the risk of confusion when configuring a custom soundbank. The commit could be applied to the AudioUnit plug-in in a similar way.
The many unused messages in this PR are copied from the JSA plugin, which has more features than the AudioUnit plugin. They can simply be removed.
Both the JSA and the AudioUnit fall back to a default sound in case the user enters an invalid soundbank file.
I have also tested the PR on macOS 14 (x86_64) with two different soundfonts and it seems to work fine.

jdemeule:
Indeed, I've copied JSA plugin message ressources and cleanup is needed.
I'm not sure to follow the point about displaying the error in case of missing/invalid sound font file, I think the last update did it.

image

Is the way of doing it does not fit you?

jdemeule:
PR updated with removing unused message and commented prototype and following what it done in jsa plugin when user enters non-existing/invalid soundbank file.

helge17:
@jdemeule, thanks for the update! I will have a look at all these warning windows in the different places in TuxGuitar and try to unify them afterwards.

I noticed one single problem: When configuring a wrong file as soundfont, e.g. a gp4 file, TuxGuitar does not start any more, but crashes with a JRE error:

```
MacOS14 tuxguitar-2025-06-19-master-macosx-swt-cocoa-x86_64.app % ./Contents/MacOS/tuxguitar.sh
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x0000000177f448c0, pid=1386, tid=26883
#
# JRE version: OpenJDK Runtime Environment Homebrew (23.0.1) (build 23.0.1)
# Java VM: OpenJDK 64-Bit Server VM Homebrew (23.0.1, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, bsd-amd64)
# Problematic frame:
# C  [CoreAudio+0xf8c0]  AUConverterFactory+0xae3
#
# No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /Users/myuser/tg-1.x-build-macos/00-Binary_Packages/tuxguitar-2025-06-19-master-macosx-swt-cocoa-x86_64.app/Contents/MacOS/hs_err_pid1386.log
#
# If you would like to submit a bug report, please visit:
#   https://github.com/Homebrew/homebrew-core/issues
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
./Contents/MacOS/tuxguitar.sh: line 24:  1386 Abort trap: 6           "${JAVA}" ${VM_ARGS} ${SWT_ARGS} -cp :${CLASSPATH} -Dtuxguitar.home.path="${TG_DIR}" -Djava.library.path="${LD_LIBRARY_PATH}" ${MAINCLASS} "$@"
```

This is a very unusual case and maybe related to my machine, but maybe you can fix that. If not, I would merge the PR anyway and open a new issue afterwards.

@guiv42, do you agree with this approach?

jdemeule:
> I noticed one single problem: When configuring a wrong file as soundfont, e.g. a gp4 file, TuxGuitar does not start any more, but crashes with a JRE error:

> This is a very unusual case and maybe related to my machine, but maybe you can fix that. If not, I would merge the PR anyway and open a new issue afterwards.
> 

This is not expected and I also reproduce it.
I will take a look at it.

guiv42:
> @guiv42, do you agree with this approach?

I don't really understand: how would it be possible to configure an invalid soundfont? My understanding is that if user selects an invalid file from the dialog, it will not be accepted. If this is correct then I can only see 2 options to reach this invalid state:
1. user has manually edited a config file. I would not mind in this case, there are already many different ways to crash the app when editing conf files manually
2. user has selected a valid soundfont once, and then replaced this valid file by another (e.g. a gp4 file). Not very probable though.

None of the 2 cases above seem critical to me, so I would consider this issue minor

Still I'm puzzled by the consequence: how can this create a crash at jvm level?

helge17:
> how would it be possible to configure an invalid soundfont?

Unfortunately, the commits 118841dd1728ec5dfcc3c8faed40b09c5b9c71be and 462d8dca16be3fed55a9e9e440adfc5d1981d27f only check if the user entered a valid file in the text field, but do not check if that file is really a soundfont, so no manual editing of the config file is needed for the crash.

Probably it would make sense to limit the file selection dialog to sf2 (and maybe sfz?) file endings, I can try to do that after this PR is merged.

> Still I'm puzzled by the consequence: how can this create a crash at jvm level?

The logfile says something like

```
# Problematic frame:
# C  [CoreAudio+0xf8c0]  AUConverterFactory+0xae3
(C=native code)
```
So I would guess the problem lies in the CPP code of the AudioUnit plugin. The JSA plugin does not crash, but reports an invalid soundfont, as expected.

guiv42:
OK, I had misunderstood, thanks for clarification (not so easy for me as I cannot run that code)

@jdemeule : would it be possible to change the way the parameter is saved?
Currently, the path to the soundfont is saved when user chooses it (parameter value is set [here](https://github.com/jdemeule/tuxguitar/blob/462d8dca16be3fed55a9e9e440adfc5d1981d27f/desktop/TuxGuitar-AudioUnit/src/app/tuxguitar/player/impl/midiport/audiounit/utils/MidiConfigUtils.java#L128) and saved a few lines after). That would be much safer if the parameter was saved only *after* the soundfont has been successfully opened (e.g. in `MidiReceiverImpl.changeSoundbank`?)
So that, if user selects an erroneous/corrupted file, it may crash the app but still enable it to restart normally after that.

jdemeule:
@guiv42, I'm doing some tests to avoid crashes.
It seems CoreAudio allow some property changes (here the valid sound bank) when the AUGraph is initialised, but starts to throw some c++ error internally on invalid properties and leave the graph in an unstable state.
Any error at this level, as the native code was loaded within the JVM address space could lead to a JVM crash (eg. SEGV).

From what I'm testing, reinitialising the graph when the sound bank changed solve the issue, report an error when the sound bank is invalid (missing, corrupt, wrong format), fallback to the default sampler (still emitting sound) and does not crash.

I will update the PR to this direction if that fit you.

helge17:
The crash has gone now, for me the PR works fine now.

guiv42:
fine for me

Issue #813, opened 2025-06-19
feature request: adapt cursor color in tracks table


guiv42:
In the tracks table (bottom part of the screen), the cursor appears as a black square. That's OK if the color associated to the selected track is bright, but if it's dark the cursor is invisible.
It would be better to adapt dynamically the color to the background (note: the background is not the same if current measure is empty or not)

guiv42:
done (#815)

PR #815, opened 2025-06-22
adapt caret color in track table according to background

guiv42:
else caret is not visible when track color is dark (#813)
also: use rounded rectangle

### DIFF
--- a/desktop/TuxGuitar/share/skins/Breeze-dark/skin.prop
+++ b/desktop/TuxGuitar/share/skins/Breeze-dark/skin.prop
@@ -1,5 +1,4 @@
 # tracks table configurable colors: R,G,B
-#table.caret=
 #table.even-line.background=
 #table.even-line.foreground=
 #table.odd-line.background=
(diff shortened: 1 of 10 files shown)


guiv42:
@helge17 : what do you think? This evolution seems necessary to me when dark colors are used for tracks, but I'm not quite sure what value to use for a threshold.
With the value I used, caret is white with the default red color for a new track.
See illustration below (image is manipulated to show all configs, of course caret can not be in 4 different measures simultaneously!)


Edit: I've got to have a more detailed look at 2.0beta fork, it seems interesting

guiv42:
Well I don't know what's best. A simple comparison between my PR and 2.0beta:

https://github.com/user-attachments/assets/17cfa97d-9eba-4a77-8250-307b4471d2c0

I'd say my PR chooses a white caret very easily, whilst 2.0beta chooses a dark one in cases where it doesn't fit really...

helge17:
I player around a little bit with different colours and would suggest to lower the brightness threshold from `3*0x80` to `3*0x50` to keep the black caret longer:

image

Then the PR is fine.

guiv42:
OK thanks, I'll do that

Issue #814, opened 2025-06-19
feature request: improve view switching


guiv42:
About (page / linear layout) and (single- / multi-track) view parameters:
- multitrack + page is unusable, it's impossible to see which line corresponds to which track
- single track + linear is just a huge waste of space

Note that this is well [documented](https://github.com/helge17/tuxguitar/blob/feddbc084b7a195a42f690abe3a550a276bc826d/desktop/TuxGuitar/share/help/detail_view.html#L77):
> good combinations are Linear+Multitrack, and Page Layout without Multitrack (for single instrument view)

So I'm switching very often between these 2 "good" configs, and it's quite cumbersome: 3-4 clicks each time, with a search of sub-items in a menu/toolbar. As many clicks to come back.
I would like very much to be able to swap in 1 single click.

Proposal, keep (page/linear) action unchanged, but modify the behavior of (multitrack on/off) actions:
- activating multitrack automatically sets page layout
- activating single track sets linear.

Interface is a checkable item (menu/button), so that should do the job: 1 single click would enable to switch between the 2 "good" configurations

To all users: what do you think?

guiv42:
This is what it would look like, with an adapted main toolbar

https://github.com/user-attachments/assets/7039dde4-853e-467e-b3c2-f0900255568a

guiv42:
done

PR #821, opened 2025-07-03
force linear layout when multitrack, page layout when single track

guiv42:
and adapt default main toolbar
see #814


guiv42:
@helge17 : what's your opinion about this PR (and associated issue)?
As a user I find it useful, and it seems to me the right moment since we are modifying the main toolbar in next release.

But as usual with user interface, there's some subjectivity here...

helge17:
I think the PR makes sense. It makes it easier to switch between the most useful layouts, but it still keeps the possibility of using the unusual layouts.

If in the future the unusual combinations turn out to be completely useless, it might also make sense to drop Multitrack altogether and just keep the Page Layout (showing only one single track) and the Linear Layout (showing all tracks) options. Or perhaps we can find a way to make the combination of Multitrack and Page Layout more useful, e.g. by coloring the notes and tablature in the same way as in the track table. But for now, I would merge this PR.

guiv42:
OK, thanks.

Would it be OK for you to update the user doc? I just had a try, but the windows decorations in my environment are slightly different than yours, that would make the screenshots not fully consistent.
I have identified the following screenshots to be updated:
- free_edition_mode.png
- layout.png
- menus_toolbars.png
- properties.png

And of course the "Layout" section of `detail_view.html` also needs an update.
If that's OK for you, you can merge this PR. Or we could also choose to do it later.

helge17:
I will care about the docs.

Issue #752, opened 2025-05-05
Apply articulation to all chord notes


rettinghaus:
Currently every note may have a specific articulation (accent, staccato, etc.), but that doesn't make sense in a chord notation-wise. If a articulation is applied to one chord note, it should also be applied to all other chord notes.

guiv42:
possible list of effects that could be moved from note to beat (please correct if needed):
- tremolo bar
- accentuated note
- heavy accentuated note
- palm mute
- staccato
- fade in

guiv42:
I'm hesitating. One the one hand this should be the right moment, since next version already breaks file format compatibility. One the other hand the new file format was merged some time ago, and has been stable since then. So I don't know.

Just a couple of things to have in mind for implementation:
- take care about old formats importers/exporters
- move controls in edit toolbar from "Effects" to "Beat" section (see pick stroke)
- same for the menu items
- some code for graphical rendering shall move from`TGNoteImpl` to `TGBeatImpl`, to avoid painting the same graphical elements several times at the same place (it's done currently, and it doesn't look very good)

Nothing's impossible, it's just a matter of available time (many other topics open currently)

belfie13:
can't you just update the other notes when an effect/articulation is set?

so if an effect/articulation is changed on a note that is in a chord, it will make the same changes to the other notes in the chord, so they don't need to be manually set.

this way the file stays the same its just the action that changes

so if you add or remove palm mute on one note in a chord, it will add or remove palm mute on the other notes in the chord for you, so we don't have to set each note manually

one thing is that you might want to have two strings muted and not the rest, so forcing the notes in a chord to have the same effects might not be a good thing

guiv42:
It's not that simple. There may be cases where one note is created/edited even before being attached to a voice/beat. Said differently, I cannot guarantee there's no such thing somewhere in the code (e.g. this could be possible in the foreign file format importers).
And there's also a graphical issue: when the same effect is applied to multiple notes in one beat it sometimes leads to a not so clean graphical result.
We'll need to decide carefully which effects need to be moved from note to beat. I don't plan to do this before 2.0 is out: the file format has been stable for a while now, I don't think it's the right moment to modify it.

Issue #262, opened 2024-03-04
1-octave shift in tuner


guiv42:
TuxGuitar 1.6.1, Linux (openSUSE tumbleweed).
Notes names displayed by tuner are 1 octave too high. A reminiscence of #37.
standard tuning is E2 A2 D3 G3 B3 E4, tuner displays E3 A3 D4 G4 B4 E5

![Screenshot_20240304_190157](https://github.com/helge17/tuxguitar/assets/129443524/91ded3ef-8f0b-4a2a-9695-8c8917b038e0)

PR #826, opened 2025-07-14
Some improvements

Daniel80MD:
Hello. I have been a user of TuxGuitar for several years. I use it for music composition and I made these changes because they usually make my work easier (although some are a matter of taste), so I think they could be useful for others as well.

These are the changes I made, for each file:

TGPiano.java:
- Improved the precision of area where the keys are pressed, now they match exactly.
- Now the notes of the scale are painted after the keys that are pressed, not being covered by them.
- A reference mark is painted on central C key.

TGMeasureImpl.java:
- If a measure has a marker, a second vertical separator bar is drawn, resulting in something like ||
- Text marker is printed at the beginning of the measure (optional using a boolean).

TGNoteImpl.java:
- Position of staccato dot was moved to near the notehead, not the stem. (Optional using a boolean).

TGGoLeftAction.java:
TGGoRightAction.java:
- Now the beat is played when the caret is moved (left or right) through it. (Optional using a boolean).

### DIFF
--- a/common/TuxGuitar-editor-utils/src/main/java/app/tuxguitar/editor/action/track/TGSetTrackInfoAction.java
+++ b/common/TuxGuitar-editor-utils/src/main/java/app/tuxguitar/editor/action/track/TGSetTrackInfoAction.java
@@ -28,7 +28,7 @@ public class TGSetTrackInfoAction extends TGActionBase {
 			TGColor color = ((TGColor) context.getAttribute(ATTRIBUTE_TRACK_COLOR));
 			Integer maxFret = (Integer) context.getAttribute(ATTRIBUTE_TRACK_MAXFRET);
 
-			if (maxFret == null) {
+			if ((maxFret == null) || track.isPercussion()) {
 				getSongManager(context).getTrackManager().changeInfo(track, name, color, offset);
 			} else {
 				getSongManager(context).getTrackManager().changeInfo(track, name, color, offset, maxFret);


Daniel80MD:
I'm sorry, I just realized that the new staccato dot position wasn't quite right aesthetically when there are multiple notes in the same measure (they all appeared in the highest or lowest note position per group). I've already fixed it, and as a "side effect", the dot now appears for each affected note (not for the entire beat).

I have updated it with a new commit on the same branch. This is my first contribution to a open source project, I hope this is the right way to do it.

guiv42:
@Daniel80MD : are you still working on this PR? I had started to review it, but it's difficult if it's not stable.

Can you please explicitly request a review (by me) when you're done?
I won't go any further until then.

Daniel80MD:
I'm sorry this was confusing., I'm not used to these processes.

No, I'm not working on this PR anymore, it's stable, and I won't touch it again unless you ask me to change something.

guiv42:
No problem, that happens. Thanks for your contribution I'll have a look.
Next time you can explicitly request a review through github (see at page top-right, you can select my name):
review

Daniel80MD:
I didn't know about review requests, thanks for the information. Of course, I'll use it next time.

guiv42:
# Review 
Many open points in this PR. A major one: this PR addresses too many different topics at once, so it's difficult to review and merge.

## Piano
I *never* used the piano in TG, so this PR is an opportunity for me to discover a few problems!
> Improved the precision of area where the keys are pressed, now they match exactly

**Question**: can you please describe the nature of the problem your PR fixes ? I don't see any major issue after playing (just a few minutes) with the piano with version 1.6.6.

> Now the notes of the scale are painted after the keys that are pressed, not being covered by them

OK, clear.

Identification of octaves on the keyboard is less obvious, and it seems [it puzzled you also](https://github.com/Daniel80MD/tuxguitar_modifications/blob/db532a7660e35f9e5d51c068ca9339ec3e29842f/desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/piano/TGPiano.java#L239). It looks very much like there is a global 1-octave shift issue (see #37, #262). TuxGuitar uses by default the *guitar treble clef*, 1 octave below *the one used for piano*. See the comparison between TG (with your PR) and [wikipedia](https://en.m.wikipedia.org/wiki/Piano_key_frequencies). Screenshots show the positions of A 440Hz, and they don't match. My interpretation is that TuxGuitar is wrong:
TG_piano
wikipedia

**Question**: do you think you can fix this? If not I will create a dedicated issue so we don't forget it.

By the way, it means there are many keys on this piano keyboard that are not usable: too low or too high. That could make sense to gray out these unusable keys, considering current track min/max pitch.
**Question**: could you address this point also? If not I'll create an issue.

And another issue: piano should *not* be usable when a percussion track is selected!
**Question**: could you address this point also? If not I'll create an issue.


## double measure bar for marker
I'm not very comfortable with this, as it highly depends from what the tab's author uses the marker for.
I would prefer to make this configurable, with default behavior unchanged (see below about config parameters)

## left alignment of marker
As you said it's for sure a matter of taste. Here again I would prefer to keep default behavior unchanged (ok to make it configurable, see below)

## Position of staccato
It's not an obvious question.
> as a "side effect", the dot now appears for each affected note (not for the entire beat).

In fact there's not much you can do here without modifying the data structure more deeply: staccato effect is associated to a *note*, even if it should probably be associated to a *voice*. See similar questions in #752. With actual TG version, if you assign staccato effect to several notes in a voice, the point is drawn several times (at the same place, so it's not visible). Anyway, I consider this is minor.
Note that moving staccato near note's head has drawbacks also:
staccato2

This is why I would prefer to keep *default* behavior unchanged.

## Play when moving left/right
Again it's a matter of taste, and again I would prefer to keep *default* behavior unchanged.

# Synthesis
Here is my proposal to go forward:
- close this PR, and re-create 1 independent PR per feature: piano / marker / staccato / play when moving
- about piano, there are a few open questions to be answered (see above) before deciding what to do next
- for the other features, I think they only have added value if the configurable options are accessible by end user (i.e. no hard-coded boolean). A possible way forward:
  - at first create new parameters, accessible through manual edition of config files. I suggest you try this with 1 single feature first, and then reproduce the behavior for the other features once it's OK.
  - then, in a second step, we could develop a UI to make the parameters available to all users. Then we shall also describe them in user doc

Don't hesitate to ask for support if needed.

guiv42 (changes requested):
see detailed review (need to break into several PRs)

Daniel80MD:
>  Many open points in this PR. A major one: this PR addresses too many different topics at once, so it's difficult to review and merge.

Yes, many open points, I did it following the directive of contribute.md which says "Take care to squash your commits before submitting a pull request". I think I misinterpreted it, it should be one commit for each point or related points.

>  Question: can you please describe the nature of the problem your PR fixes ?

The problem is that when I click on the red box (in the following pic), the sharp key (cyan) is activated instead of the natural key (which is what I expect since it is the one I am pressing). For someone who uses the piano a lot, this is frustrating.
piano

>  Identification of octaves on the keyboard is less obvious.

Well, my original idea was mainly to be able to visually move more quickly between such a long row of similar keys, since I use the piano a lot for composition. I thought the most logical thing was to make it matches the matrix editor, to always know "where you are" in both editors. But any other representation is also fine.

> Question: do you think you can fix this? If not I will create a dedicated issue so we don't forget it.

Sorry, I don't feel able to modify the piano layout in the short term.

> And another issue: piano should not be usable when a percussion track is selected!

In fact, I use it a lot for drum tracks because it seems to respect the standard for drum keys in digital pianos (I have had several physical digital pianos and the keys match, and I have more or less memorized where each drum instrument is on the piano keyboard). I had even thought about putting the name (or something else like a icon) of the drum instrument on each key in some visual way, but the space is too small. But if you want to disable the piano for drum, it is also OK :-)

> I'm not very comfortable with this, as it highly depends from what the tab's author uses the marker for.

I agree, I would prefer a standalone way to add a separator without using a marker (and without carrying a marker when I don't need it), but I wasn't able to do it. Well, I think my solution was not very good.

> Position of staccato

I agree. I just read the theory the theory on Wikipedia, it seems like a more complex issue, yes, I agree to leave it as it is  currently. It should be noted that although only one staccato point is painted (several in the same place), it affects per note, not per beat (but you can see the button pressed for which note has the staccato, so it's not a problem).

> Play when moving left/right

I use this feature intensively while composing, In one of my local versions I even put a button (from the PC-keyboard, not in the GUI) to play the beat at any time without moving (this speeds up my "testing/iterations" of the music). But I understand that maybe other people are bothered by hearing the beat every time they move, that's why I made it easily configurable using a boolean (I agree that this should be in an configuration dialog, not hardcoded). If you decide it shouldn't be included, that's also fine, I was just suggesting it :-)


> Synthesis

I also think that features should be configured by the end user, I totally agree, but I didn't dare to change too many things. My main idea was to show in code where those new features could start (instead of just asking for them), and if my code is suitable, even better.

I think it's a good idea to close it and make a PR for each feature you want to include. I could try adding the configurable features to TGConfigKeys (or wherever you tell me) and put them for end-user in this already-existing menu:
opciones_existentes

My list of new PRs would be:

1. Piano: fix the area of the keys, and notes of the scale painted after the keys that are pressed. Identification of octaves: discarded (because that has to be decided better).
2. Marker double bar and position: both configurable in options menu (later could be in marker itself or there could be independent separators, but that would mean changing the file format and I don't feel ready for that), or discard it so that perhaps in the future it will be addressed in a better way (whatever you prefer).
3. Play when moving: configurable in options menu.

Discarded features:
- New staccato dot position (In fact, I really don't know why I got into complication with that).

guiv42:
Thanks for all your answers.

The recommendation to squash commits is just to avoid many intermediate dev commits with low added value (like modifying 10 times the same file until it stabilizes).

**Piano**
OK for the fix about clicking on the keyboard, it's clear.
About the identification of octaves, I propose to keep your implementation unchanged. There is for sure a 1-octave shift, but it's been there forever and at least your implementation makes it visible. I will create a dedicated issue so we don't forget this point, it can be fixed later.

Using piano keyboard to manipulate percussion track is very suprising to me! But anyway, if it can be helpful in some manner, then there's no reason to change this.

**Double measure bar**
In fact it's probably easier (and cleaner) to add a new feature, independent from markers. It could be a new item in "Measure" menu, and/or an icon in toolbar(s). That's just a new attribute to add in `TGMeasure` (and an optional field in .tg file format), nothing too complex I think.

**Staccato**
I agree with you, there's a significant potential for improvement, but that also means a significant workload!

**Play when moving**
I understand your point, it clearly depends how everyone uses TG. In this case, both behaviors make sense I think, so making it configurable seems the right method to me.
To add a new parameter, you need to define it in [`TGConfigKeys`](https://github.com/helge17/tuxguitar/blob/master/desktop/TuxGuitar/src/app/tuxguitar/app/system/config/TGConfigKeys.java) and to define the default value in [`TGConfigDefaults`](https://github.com/helge17/tuxguitar/blob/master/desktop/TuxGuitar/src/app/tuxguitar/app/system/config/TGConfigDefaults.java). Then, to use it you need to retrieve an instance of [`TGConfigManager`](https://github.com/helge17/tuxguitar/blob/master/desktop/TuxGuitar/src/app/tuxguitar/app/system/config/TGConfigManager.java) from a `TGContext` instance (that may be the only tricky point, depending where you want to access the parameters). Ask for support if needed.
As I mentioned earlier, you can do it in 2 steps: first create the parameter and use it, then we can modify the UI.


This leads me to the following **suggestions**:
1. piano: keep your implementation unchanged, *including octaves numbering*
2. double bar: make it an independent feature. If you don't feel capable of implementing it, could you please create a new issue ("feature request: double bar")? I can probably implement that.
3. marker position: make it configurable
4. play when moving: make it configurable

Daniel80MD:
I agree with the suggestions. I'm going to go step by step. 
I'm also learning how the GitHub system works.

PR #832, opened 2025-07-18
Piano fixes: click area and drawing order

Daniel80MD:
This is a pull request only for the piano fixes:
- Click area fixed.
- Drawing order fixed (this simply changes the place of line 247).

This does not include the identification of octaves.


Daniel80MD:
I don't see any options to explicitly request a review:
REVIEW
I guess it doesn't have to appear yet.

guiv42:
> I don't see any options to explicitly request a review

It should be there... Here's what I see on another PR, if you see this you can click on the icon on the right and select a reviewer.
review

Not a problem anyhow.

guiv42:
I think I got confused by github interface: I have already received spontaneous requests for review from contributors, but they were *2nd* requests (after a 1st review)...

PR #848, opened 2025-08-04
GUI for play-when-move with two languages

Daniel80MD:
I have made a GUI for play-when-moving, in case you want to add this feature to the official version.

I have added English and Spanish (because Spanish is my native language). I could try to translate the other languages using Artificial Intelligence, although I assume you already have a better way to do it.

Hope it helps.

configure_con_playwhenmove

### DIFF
--- a/common/resources/lang/messages.properties
+++ b/common/resources/lang/messages.properties
@@ -705,6 +705,7 @@ settings.config.main.window-title.var.description.songauthor=This variable repre
 settings.config.main.options=Options
 settings.config.main.table.auto-size.enabled=Enable Automatic Table resizing
 settings.config.main.splash-enabled=Launch splash screen on startup
+settings.config.main.play-when-moving=Play the beat when moving the cursor
 settings.config.styles=Styles
 settings.config.styles.general=General Styles
 settings.config.styles.printer=Printer Styles
--- a/common/resources/lang/messages_es.properties
+++ b/common/resources/lang/messages_es.properties
@@ -562,6 +562,7 @@ settings.config.main.window-title.var.description.songauthor=Representa el autor
 settings.config.main.options=Opciones
 settings.config.main.table.auto-size.enabled=Ajustar automáticamente el tamaño de la previsualización de pistas
 settings.config.main.splash-enabled=Mostrar el splash al inicio.
+settings.config.main.play-when-moving=Reproducir el tiempo al mover el cursor
 settings.config.styles=Estilos
 settings.config.styles.general=Estilos Generales
 settings.config.styles.printer=Estilos de Impresion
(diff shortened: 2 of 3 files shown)


helge17:
A GUI element would certainly be useful, but I think it would fit better in the main menu, perhaps at the end of the “Player” menu under “Highlight Played Beat”? A keyboard shortcut could also be defined for this.

Pruchniakowski:
Polish (and other Slavic languages) translated by AI? Sounds good. There will be a lot of laughter! I approve this idea!

guiv42:
> I have added English and Spanish (because Spanish is my native language)

@Daniel80MD : up to now, all translations have been produced by the community, and I see no reason to change this. So what you did is OK.

> I think it would fit better in the main menu, perhaps at the end of the “Player” menu under “Highlight Played Beat”? A keyboard shortcut could also be defined for this.

@helge17 : I don't know. The "Settings" window makes sense, as it is clearly a way to configure the behavior of TG. I had initially thought of this window, but in the "Sound" section. I see no added value to enable a keyboard shortcut for an action which should be used very rarely (subjective opinion, of course).

@Pruchniakowski : I agree with you, but please change your attitude. Constructive criticism will always be accepted, but irony and sarcasm have no place here.

Pruchniakowski:
@guiv42 Are you typing about yourself?

Pruchniakowski:
I have a question: helge17 and guiv42 is/are the same person?

Pruchniakowski:
Pathetic...

guiv42:
@Pruchniakowski please stop your provocations. In the future I will delete all your inappropriate comments.

Pruchniakowski:
You scared me 😆

Daniel80MD:
@guiv42 : I'm glad the community does the translations, it's better that way.

I also think it's a configuration option that is changed very rarely, so a shortcut wouldn't be necessary.

If you decide that I should put that option in another place, please let me know.

helge17:
> The "Settings" window makes sense, as it is clearly a way to configure the behavior of TG. I had initially thought of this window, but in the "Sound" section. I see no added value to enable a keyboard shortcut for an action which should be used very rarely (subjective opinion, of course).

The “Settings” window imo configures the general and “low level” behavior of TG. Since the play-when-move feature is related to the playback of sounds, I thought it would be better to place it under the "Player" main menu. But that's a matter of taste, so I'm also fine with adding it to the “Settings” window.

I agree that a shortcut does not add much value, so we can of course leave it out.

I have also considered using AI for translations, e.g. for the help pages. However, without at least a basic knowledge of a language, I consider it too risky to accept these translations without human verification.

guiv42:
Windows have a significant advantage over menus: there is more available space (width) so it's possible to write more extensive descriptions.
A description like "Reproducir el tiempo al mover el cursor" would probably appear quite large in a menu, and other languages may need even more space to be clear. Menu items need to be concise and brief, making translations more difficult I think.
So I would vote to keep the option where it is currently.

helge17:
>  So I would vote to keep the option where it is currently.

That's also fine for me. @guiv42, just merge the PR when you think its ready.

guiv42 (approved):

guiv42:
@Daniel80MD thank you for your contributions.
Do not hesitate to edit [AUTHORS](https://github.com/helge17/tuxguitar/blob/master/AUTHORS) file.

Daniel80MD:
@guiv42  Thank you! I enjoy contributing.

Issue #835, opened 2025-07-19
feature request: shift piano by one octave


guiv42:
Piano displayed by TG shows notes from A(-1) to B6.
According to [wikipedia](https://en.m.wikipedia.org/wiki/Piano#Keyboard) a standard 88-key piano covers interval from A0 to C8.
As a consequence, the "middle C" corresponding to C4 does not appear in the middle of the keyboard, but one octave at the right.

It would be interesting to:
- shift displayed range by 1 octave: from A0 to B7
- grey out the many keys that are unusable, considering current track min/max pitch. For example the guitar track in default template covers range E2 to A6

See discussion in #826

Issue #131, opened 2023-10-06
Extended percussion instruments missing from the matrix editor


buryoutougen:
There are some percussion sounds available below note 35 and above 81, but you can't input them using the matrix editor. You need to use an external reference like [this](https://en.wikipedia.org/wiki/General_MIDI#GS_extensions) or go by trial and error.

guiv42:
Reading the wikipedia article you mentioned, I understand that only notes 35 to 81 are part of "GM standard MIDI". Outside of this range:

> Roland GS is a superset of the General MIDI standard that added several proprietary extensions

So I'm not sure it's a good idea to add these **proprietary** extensions, as users may export MIDI files from TuxGuitar to play them with other devices than Roland. Depending from the MIDI synthetizer used, I'm not even sure TuxGuitar would be able to play these sounds.
In other words, I would prefer to keep actual behavior unchanged.
Is my interpretation of this article correct?

buryoutougen:
Sorry, maybe I wasn't completely clear, or I don't fully understand what the problem is.

These sounds are already in TuxGuitar, it's just that you have to input the note numbers manually, as they can't be selected from the matrix editor.

I'm not an expert on MIDI either, but my understanding is that these additional notes were added by Roland and then standardized in General MIDI 2 (also see [here](https://soundprogramming.net/file-formats/general-midi-level-2/) and [here](https://soundprogramming.net/file-formats/general-midi-drum-note-numbers/)).

Anyway, it's not a big deal, I just thought it's something that could be improved, as someone might not be aware of these additional sounds if they only use the matrix editor to input drums.

guiv42:
OK, got it, thanks for your explanation.
This should be easy to add, pull request coming soon.
Note: all these sound names are hard-coded in English, that might be interesting to open them for translation one day...

guiv42:
Is it just my configuration, or is there a problem of sound within the matrix editor?
When I click in one cell of the matrix, it produces the expected sound. If, then, I click in another cell, the sound I hear corresponds to my first click, not the second instrument... Clicking twice in the second cell (remove / add sound again) and it produces the correct sound.

buryoutougen:
Yes, same for me. It seems to me that when you click on a cell, TuxGuitar first plays the notes from the currently focused column, then switches focus to the column you clicked - should be the other way around.

guiv42:
OK, I understood the cause of that problem, it should be fixed by pull request #134.
[...]

PR #855, opened 2025-08-08
shift piano by one octave and show non-playable area

guiv42:
see #835


guiv42:
@Daniel80MD I would appreciate your opinion about this PR.
Especially about the representation of non-playable area. It is automatically adapted when selecting a track with another tuning. Nothing special is shown for percussion tracks.
Here is what it looks like with a standard tuned guitar:
piano

Daniel80MD:
@guiv42 
Hi, I compiled this PR and tested everything overall. I noticed some minor inconsistencies. I'm not entirely sure if this is exactly what you wanted me to focus on, but based on my understanding, here are my observations.

With a standard tuned guitar (with 29 frets), in a previous version (not this PR) I could play up to B6 (red square), and the left limit is the same as in this PR:
GUITARRA_PLAYABLE_EN_ANTERIOR_VERSION
However, although the instrument definition includes 29 frets, the fretboard in the GUI only shows 23 frets, so on the fretboard I can only play up to D#6 (blue square on piano). It's a minor inconsistency but nothing to worry about.
MASTIL_GUITARRA_DS

---

And if this is useful: for a percussion track, in your PR version (not the previous version), the playable instruments would be from 27 (High Q, D#1) to 87 (Open Surdo, D#6). I assume there are no other General MIDI drum instruments outside that range, so I assume it is safe to disable those keys:
BATERIA_PLAYABLE

guiv42:
Thank you @Daniel80MD for your answer.
The discrepancy between track fret number (default 29 frets) and fretboard representation (23 frets) has been there forever as far as I know. The only clean solution would be to make the fretboard size user configurable.
For percussion tracks: the min/max limitation to apply is not so clear, it seems there are different standards (e.g. see discussion in #131). So I prefer to leave the current behavior unchanged. You can already enter "notes" outside of the [27..87] range with the keyboard, so I see no reason to limit the range of the piano.

Daniel80MD:
@guiv42 I imagine making the fretboard configurable could be quite complex.

I didn’t know that about percussion and wasn’t really sure. Thanks for letting me know. I agree that making no changes to percussion is the best approach.

guiv42:
> I imagine making the fretboard configurable could be quite complex

I don't think so. Most probably it's just a hard-coded [constant](https://github.com/helge17/tuxguitar/blob/da24e0c49705d5065fbcdf9cc8e8f5a6a173bbc3/desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/fretboard/TGFretBoard.java#L60) to replace by a user-defined parameter. Just need to find some place in the GUI for a spinner.
I'll probably do that one day, but currently I would like the code base to stabilize, to enable the publication of next official release.

PR #858, opened 2025-08-08
help: minor precisions

guiv42:
Thanks @helge17 for the update of user help.
Here are a few proposals for precisions


helge17:
Thanks!

Issue #885, opened 2025-09-14
Bug: Play mode -> Training mode -> "To" scroll box is too restrictive


ElektroVirus:
The _To_ scroll box is too restrictive and makes it almost impossible to write the wanted percentage by keyboard, requiring to use the + / - buttons.

Example: write 50 to the _Tempo_ scroll box. Press tab, write 60 to the _To_ scroll box. The scroll box contains now text 051. Apparently it thinks that the first character 6 is less than 50 in the _Tempo_ scroll box, which makes it illegal. At least when the _Tempo_ scroll-box is 5%, the _To_ scroll box accepts both 6 and 60.

<img width="508" height="625" alt="Image" src="https://github.com/user-attachments/assets/75e0625b-ff6c-465e-a251-8ec3a028d4b8" />

Software:
- TuxGuitar 2.0.0beta1 (swt)
- Bug is  also present on TuxGuitar 1.6.6_linux-swt
- Fedora Linux 44

guiv42:
It's not so obvious, as it seems the ui framework (SWT) does not handle a timer to distinguish user input between "60" and "6  then 0".
One possible workaround could be to accept all user inputs (including "6"), but to make "OK" button enabled only when value of "To" is greater than "From".

For information, this problem is pretty old (also present in historic version 1.5.6), so I consider it low priority: to be fixed *after* 2.0 is out (we're currently in a code stabilization phase)

ElektroVirus:
I totally agree, this is not on top of priority list. Just wanted to list this issue, to help improving the usability of TuxGuitar piece by piece.

I also thought about possible solutions but none were great. One option would just be to accept every number in the scroll box and if _To_ would be smaller than _Tempo_, just set _To_ as _From + 1_ silently. But I'm not a fan of silent changes where the user is not notified.

Your proposed solution sounds better. Also some kind of label where an error message explains why the Ok-button is disabled, would be helpful.

guiv42:
> Just wanted to list this issue, to help improving the usability of TuxGuitar piece by piece.

Thanks for reporting, it clearly makes sense.
I share your opinion about silent changes, and also about the need to inform user that something is wrong. But I don't really like the idea of an error message: when user types "60", that would make the error message appear after the "6" and disappear immediately after the "0". So it may just *blink* for a short time.
Another possible option could be to change the color of the scroll-box (either foreground or background). That could be worth a try.

guiv42:
[...]

PR #896, opened 2025-09-22
player mode dialog, simplify entry with keyboard

guiv42:
see #885
Scenario:
- open player mode dialog (F9)
- activate "Training mode" option
- select "From" and type "50"
- type "tab" key (it selects "To"), and try to type "60"

There is no timer to distinguish between "60" and "6 ... then 0"
Before this PR, the "6" was ignored because 6<50, so it's an invalid value

This PR applies another approach: all user inputs are accepted but if From/To are inconsistent, OK button is disabled and last edited field is shown with red background

### DIFF
--- a/desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/transport/TGTransportModeDialog.java
+++ b/desktop/TuxGuitar/src/app/tuxguitar/app/view/dialog/transport/TGTransportModeDialog.java
@@ -349,34 +349,40 @@ public class TGTransportModeDialog {
 		}
 
 		public void onSelect(UISelectionEvent event) {
-			if( event.getComponent().equals(this.from)){
+			boolean ok = true;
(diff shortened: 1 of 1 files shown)


guiv42:
@helge17 if you have some time, I would like your opinion about this user interface evolution (and eventually the new color). See screenshots and short video in #885

guiv42:
OK, thanks, I'll have a try

guiv42:
This should be OK now:

https://github.com/user-attachments/assets/e73b2791-2b65-406f-9e5c-3d47c57f7419

helge17:
Perfect, thanks! Just merge it if you (and @ElektroVirus?) think it's ready.

ElektroVirus:
Splendid! I really like this improvement. Thank you so much for the hard work!

I  also Commented to the original issue:
https://github.com/helge17/tuxguitar/issues/885#issuecomment-3337406873

Issue #895, opened 2025-09-21
macOS Dark Mode Detection


belfie13:
Currently the selected skin is used when dark mode is enabled on macOS. this wouldn't be a problem however the side bars, toolbars, and other accessory views are all darkend by the system and only really works if you switch to the dark version of the skin.

the feature request:

set tuxguitar skin to dark dependant on macOS dark mode setting
- detect macOS dark mode setting on startup
- automatically switch to dark skins when dark mode is enabled in macOS.

might be worth noting, sublimetext has settings for `color_scheme`, `light_color_scheme`, and `dark_color_scheme`. the `color_scheme` setting can be set to `auto`, that switches between the light/dark scheme with the operating system.

so if you just set the `color_scheme` then it's static but if you set `color_scheme=auto` then it uses `color_scheme=light_color_scheme` or `color_scheme=dark_color_scheme` if dark mode is enabled in macOS.

i assume there's an NSNotification you can register for in macOS but it's been a decade since i programmed for apple devices.

v2.0.0 beta2 macosx swt cocoa x86_64

belfie13:
[https://github.com/helge17/tuxguitar/issues/878#issue-3385016187](url) related

guiv42:
Detection of operating system dark mode was also suggested [here](https://github.com/helge17/tuxguitar/issues/27#issuecomment-2435900528). What makes it difficult is that it's fully OS-dependent, so that means at least 3 different implementations.
Another significant problem is I cannot access to any apple device, so I am not able by myself to try to implement anything (note: I have the same problem with windows 11)

The good news is that 2.0.0 removed some old skins that were not maintained and did not have light/dark variants. Now the 3 available skins are available with this option, so if we succeed to detect OS dark mode we should be able to switch skin automatically in all cases.

A plugin is most probably the easiest way to handle the dependency to OS (we could have 1 specific plugin for each OS).
If someone is motivated, there could be some inspiration in this project: https://github.com/Dansoftowner/jSystemThemeDetector

guiv42:
I tried an implementation, can someone with a macOS device test my [branch](https://github.com/guiv42/tuxguitar/tree/darkModeMacOS)?
There should be binaries available [here](https://github.com/guiv42/tuxguitar/actions/runs/18358058833) thanks to github actions (I don't know how long these binaries remain available)

[...]

PR #920, opened 2025-10-10
Detect dark mode when app starts (macOS, Linux/KDE)

guiv42:
see #895

@helge17, just one open question:
I have implemented a user parameter to disable the feature, in case there is a problem in some specific configurations. But I did not add any user interface for this: let's say I'm optimistic, and I hope it won't be necessary. What do you think?

### DIFF
--- a/desktop/build-scripts/common-resources/common-linux/tuxguitar.sh
+++ b/desktop/build-scripts/common-resources/common-linux/tuxguitar.sh
@@ -1,4 +1,16 @@
 #!/bin/sh
+## try to read light/dark theme from dbus
+getTheme() {
+    tuxguitarTheme="none"
+    # does not work with all desktops, limited to KDE
+    if [ $XDG_CURRENT_DESKTOP = 'KDE' ]; then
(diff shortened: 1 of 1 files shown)


G-eos:
Works for me on MacOS 15.6.1 thanks :)

guiv42:
For information, I tried another approach in order to be more generic (this PR does not work in all Linux configurations):
- when app starts, retrieve app main window background
- modify selected skin accordingly

The objective was to be more robust. Practically it's not so important to retrieve light/dark mode *applied by the OS*. What is really important is to retrieve the mode *applied by TuxGuitar* (and that depends from external lib SWT/JFX)

Unfortunately it does not work: skin manager is initialized *before* the main window

helge17:
> @helge17, just one open question: I have implemented a user parameter to disable the feature, in case there is a problem in some specific configurations. But I did not add any user interface for this: let's say I'm optimistic, and I hope it won't be necessary. What do you think?

I think it's better to omit the user interface. After your commit, I will add a note about this feature to the documentation.

guiv42:
switching PR to "draft"
Still working on another solution, supposed to be more stable: don't detect the theme *applied by the OS*, but try to detect the theme *applied by TuxGuitar*.

guiv42:
re-opening PR, since #924 does not work properly on macOS

PR #924, opened 2025-10-19
detect dark mode applied by TuxGuitar when app starts and adapt skin 

guiv42:
a user config parameter enables inhibition of the feature

This is an attempt to supersede #920.
Comparison with #920:
- it is now consistent with the light/dark theme applied by SWT (not yet tested with JFX)
- code is independent from OS and desktop (KDE/Gnome/...)

@helge17 and/or @G-eos: could you please test on macOS?

### DIFF
--- a/desktop/TuxGuitar/src/app/tuxguitar/app/system/config/TGConfigDefaults.java
+++ b/desktop/TuxGuitar/src/app/tuxguitar/app/system/config/TGConfigDefaults.java
@@ -33,6 +33,7 @@ public class TGConfigDefaults{
 
 	public static void loadProperties(TGProperties properties){
 		loadProperty(properties, TGConfigKeys.SKIN, DEFAULT_SKIN);
+		loadProperty(properties, TGConfigKeys.SKIN_DARK_AUTO, true);
 		loadProperty(properties, TGConfigKeys.WINDOW_TITLE, "${appname} - ${filename}");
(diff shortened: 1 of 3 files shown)


helge17:
On Debian 12 with SWT, the light/dark theme is applied automatically (KDE).
On Debian 12 with JFX, the theme does not change automatically. But the menues remain bright anyway, even if KDE is set to dark mode.

On macOS, light/dark mode detection does not work (SWT & JFX), you must switch modes manually. TuxGuitar always uses light mode when it starts, regardless of how the desktop is set or what was previously set manually.

My test with SWT on macOS:

- OS: Set desktop to light mode
- OS: % rm -r ~/Library/Application\ Support/tuxguitar
- TG: start
- TG: Breeze applied
- TG: exit
- OS: Switch desktop to dark mode
- TG: start
- TG: Breeze applied (not Breeze Dark)
- TG: Switch theme to Breeze Dark manually
- TG: exit
- OS: Switch desktop to bright mode
- TG: start
- TG: Breeze applied (not Breeze Dark!)
- TG: Switch theme to Breeze Dark manually
- TG: exit
- TG: start
- TG: Breeze applied (not Breeze Dark!)

guiv42:
OK, thanks for the test. Init sequence of SWT objects possibly differs from one OS to another... 
So let us come back to #920

PR #939, opened 2025-11-06
Portuguese Translation Improvements

LeonardoQF:
These changes aim to improve the already present Portuguese translations, and add them where they were not present.

Biggest changes include:

- Changed the translation of track from "Pista" to "Faixa" ("Faixa" is more fit for music, where as "Pista" is for, say, racing tracks);
- Translated midi instruments' and midi percussion instruments' names;
- Fixed typos and updated some existing translations to be more seamless and compliant with the English version;

- Added Portuguese translation files and translations to:
 - Tuner;
 - Batch converter;
 - Jack File;
 - PDF Export tool.
 
 The translations have been tested and seem to be working without any integration problems.


nbkhope:
Is there a distinction between Portuguese from Brazil (pt-BR) and Portuguese from Portugal (pt) in the TuxGuitar internationalization (i18n)?

common/resources/lang/messages_pt.properties
```diff
@@ -249,76 +249,76 @@ track.properties.general=Geral
 track.color=Cor
 track.instrument=Instrumento
 # track.instrument.default-select-option=
-# track.maxFret=
-# track.confirm.reduce-fret-number=
+track.maxFret=Número da Fret
```
nbkhope: Should _casa_ be used instead of fret?

common/resources/lang/messages_pt.properties
```diff
@@ -249,76 +249,76 @@ track.properties.general=Geral
...
+note.shift-down=Mover uma Corda Abaixo
+beat.move-left=Mover um tempo para a Esquerda
+beat.move-right=Mover um tempo para a Direita
 beat.move-custom=Mover tempo Personalizado
 beat.move-custom.dialog.title=Mover Personalizado
 beat.move-custom.dialog.direction-tip=Direção do Movimento
 beat.move-custom.dialog.direction=Direção
-beat.move-custom.dialog.direction.right=Mover para Direita
-beat.move-custom.dialog.direction.left=Mover para Esquerda
+beat.move-custom.dialog.direction.right=Mover para a Direita
```
nbkhope: Maybe the intent is to be brief and omit the `a`

LeonardoQF:
> Is there a distinction between Portuguese from Brazil (pt-BR) and Portuguese from Portugal (pt) in the TuxGuitar internationalization (i18n)?

There is not, there are only Portuguese translation files (pt) and they are used for both. I think it works, as they are quite similar.

common/resources/lang/messages_pt.properties
LeonardoQF: I thought the same, even though using "a" before directions is the grammatically correct of saying it, it comes at the expense of conciseness.
  Would it be better to ditch the "a" for that sake?

common/resources/lang/messages_pt.properties
```diff
@@ -683,287 +683,287 @@ action.measure.go-first=Ir para o primeiro compasso
 action.measure.go-last=Ir para o último compasso
 action.measure.go-next=Ir para o próximo compasso
 action.measure.go-previous=Ir para o compasso anterior
-# action.measure.fix-voice=
+action.measure.fix-voice=Fixar voz
```
LeonardoQF: By the way - In the case of this line, is _Fix voice_ in the sense of "repair" or "pinning"?
The translation to "fix" changes completely depending on what it is.

helge17:
Thanks!

Issue #887, opened 2025-09-15
Forward autoscrolling


bergentroll:
When playing "from the list" it is expected to see few bars forward. Current TuxGuitar scrolls screen-by-screen on the very last note of the last bar in the viewport while it is playing.

It is supposed to be more convenient to leave 1-2 bars from the next screen in the viewport while playing both for the linear mode and the page mode. For sure it is good to have an integer option for the forwarding bars starting from `0`.

Forwarding bars may looks e.g. half-transparent to not to mislead a user.

It may be also nice to have continuous mode with fluent scrolling, but suppose it is hard to render.

guiv42:
I made a quick & dirty prototype of continuous scrolling just to see what it could look like. My first conclusions:
- most probably **it will not work correctly in linear mode**. When measures are wide and screen isn't, required scrolling speed makes the tablature not readable at all. Discontinuous scrolling (a few measures in anticipation) is probably the best solution
- in page mode, vertical scrolling needs to be much slower, this could possibly work

A key parameter is to evaluate a **max scrolling speed** (pixels/second) to keep tablature readable. Then, when this max speed is not fast enough to enable continuous scrolling it is detectable, and discontinuous scrolling can be applied

If someone has ideas about how to evaluate this max speed, please reach out!

Following videos need watching in full screen
Prototyping X-scrolling:

https://github.com/user-attachments/assets/f34860af-7249-44ca-a5f2-08fa13712f02

Prototyping Y-scrolling (same song, same window size)

https://github.com/user-attachments/assets/159b5fa5-6cad-4df6-97ee-143123c8b435

bergentroll:
Vertical continuous scrolling seems really not bad! Some thoughts:

- For vertical mode at end of some lines there is a pause on scrolling. It may looks better if speed will be constant till the end of a track.
- Horizontal continuous scrolling may looks better if play marker will be fixed not far from the left side of the viewport while scrolling goes. So eyes will not travel a lot by the line.
- The continuous scrolling may be nicely used as an FX for manual scrolling. It may gives smooth "natural" scrolling experience which is good for perception.

Thx!

guiv42:
> It may looks better if speed will be constant

Constant scrolling speed is not a reasonable objective. E.g. some lines may hold 2 measures only, others can hold 6 or more, so the scrolling speed needs permanent adaptation.
I am aware of the limitations you mention. Don't get me wrong: this is a quick & dirty prototype, I'm very far from a working implementation.

[...]

PR #942, opened 2025-11-10
continuous scrolling when playing

guiv42:
see #887

- continuous scrolling: inactive by default resource consuming, even after some optimization performance issues can remain visible.
still TODO: icon (player menu and main toolbar item)

also:
- discrete scrolling: anticipation by 2 measures.

@helge17 I would appreciate your opinion on this. Not on code review, it's more about a subjective perception of the feature.
- horizontal scrolling is practically unusable, it's too fast to be readable (not an implementation problem)
- vertical scrolling: on a low-performance hw config there's a lot of "flickering"

I spent some time optimizing, but I don't have any other idea to go further.
Can this feature still be useful, even if not all users can use it?

### DIFF
--- a/desktop/TuxGuitar/src/app/tuxguitar/app/system/config/TGConfigDefaults.java
+++ b/desktop/TuxGuitar/src/app/tuxguitar/app/system/config/TGConfigDefaults.java
@@ -154,7 +154,8 @@ public class TGConfigDefaults{
 		loadProperty(properties, TGConfigKeys.CHORD_INSERT_DIAGRAM_ONLY, false);
 
 		loadProperty(properties, TGConfigKeys.SCROLLING_MAX_FPS, 40);
-		loadProperty(properties, TGConfigKeys.SCROLLING_DISCRETE_ANTICIPATION, 1);
+		loadProperty(properties, TGConfigKeys.SCROLLING_HORIZONTAL_DISCRETE_ANTICIPATION, 1);
+		loadProperty(properties, TGConfigKeys.SCROLLING_VERTICAL_DISCRETE_ANTICIPATION, 1);
 		loadProperty(properties, TGConfigKeys.SCROLLING_HORIZONTAL_MARGIN_PERCENT, 20);
 		loadProperty(properties, TGConfigKeys.SCROLLING_VERTICAL_MARGIN_PERCENT, 0);
 	}
--- a/desktop/TuxGuitar/src/app/tuxguitar/app/system/config/TGConfigKeys.java
+++ b/desktop/TuxGuitar/src/app/tuxguitar/app/system/config/TGConfigKeys.java
@@ -128,5 +128,6 @@ public class TGConfigKeys {
 	public static final String SCROLLING_MAX_FPS = "scrolling.continuous.maxFPS";
 	public static final String SCROLLING_HORIZONTAL_MARGIN_PERCENT = "scrolling.continuous.horizontal.margin-percent";
 	public static final String SCROLLING_VERTICAL_MARGIN_PERCENT = "scrolling.continuous.vertical.margin-percent";
-	public static final String SCROLLING_DISCRETE_ANTICIPATION = "scrolling.discrete.anticipation";
+	public static final String SCROLLING_HORIZONTAL_DISCRETE_ANTICIPATION = "scrolling.discrete.horizontal.anticipation";
+	public static final String SCROLLING_VERTICAL_DISCRETE_ANTICIPATION = "scrolling.discrete.vertical.anticipation";
 }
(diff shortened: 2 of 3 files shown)


guiv42:
Thank you very much for your feedback.
**horizontal continuous scrolling** : I'll have a try with a configurable margin at the left
**anticipated discrete scrolling**: if you want to test, you can change the number of anticipated measures in the config file. That may even be useful to give user the possibility to set this parameter, but I don't really know where in the user interface. Any idea?

helge17:
> anticipated discrete scrolling: if you want to test, you can change the number of anticipated measures in the config file.

I played around with scrolling.discrete.nb-measures-anticipation and noticed the following:

Setting scrolling.discrete.nb-measures-anticipation=[432] works fine: The cursor jumps to the top when there are [432] measures left to play, exactly on the last beat of the [543]th last mesure.

scrolling.discrete.nb-measures-anticipation=1 is wrong: The cursor jumps back at the end of the **last** visible measure. It should jump back at the end of the previous measure when there is one single full measure left to play. So it behaves like scrolling.discrete.nb-measures-anticipation=0.

scrolling.discrete.nb-measures-anticipation=0 (and also ...=-1) is fine: The cursor jumps back at the very last beat of the last visible measure.

> That may even be useful to give user the possibility to set this parameter, but I don't really know where in the user interface. Any idea?

I'm not sure if such details should be configurable via the UI. I think a reasonable default setting (1 or 2?) is fine for >> 90% of users; too detailed settings will probably confuse most users. And if someone really wants to change this, they can do so in the configuration file.

guiv42:
> scrolling.discrete.nb-measures-anticipation=1 is wrong

Confirmed (reproduced), I'll fix that.

OK for the default setting without specific UI. Thanks for your feedback

guiv42:
Done in the last 2 commits:
- fix discrete scrolling with 1 measure anticipation
- set default value for discrete scrolling nb measures anticipation to 1
- continuous horizontal scrolling: it now keeps a margin at the left of 20% of the display area width

@helge17: what do you think?
I'm not 100% convinced by the left margin when scrolling horizontally. When measures are wide, it increases the risk to trigger a discrete scroll, when it cannot scroll fast enough (practically, usable area is reduced by 20%)
If the behavior is acceptable for you, I could as well move this hard-coded 20% value in the user configuration file.

guiv42:
39f6711531dca9f1857411aa08641874a82712d7 is my first attempt to define a vector icon, I hope I have done it correctly.
@helge17 feel free to modify anything if needed (including the icon itself)

helge17:
Discrete scrolling:

scrolling.discrete.nb-measures-anticipation=1 now works perfectly, thanks!

In page layout, an anticipation > 0 has one drawback: the cursor does not jump to the upper left corner of the window, but jumps to the next measure in the first line and shortly thereafter to the second line, which I find somewhat confusing.
In linear layout, the behavior is fine; the cursor always jumps to the left edge of the window.

Perhaps the meaning of anticipation=X should be changed: in **page layout**, scrolling occurs when X **lines** remain, in **linear layout**, scrolling occurs when X **measures** remain; default value X=1. But that's a matter of taste, maybe it becomes too complicated that way.

Continuous scrolling:

Linear layout, 20% left margin: For a song without repeats, the left margin may seem wasted, that's true. However, since most desktop screens are in landscape format, I don't think the 20% waste is that much of a problem. If you have a lot of short repeats, such as in [acdc-girls_got_rhythm.gp3](https://github.com/user-attachments/files/23465290/acdc-girls_got_rhythm.zip) from #943, the 20% left margin avoids discrete scrolling back, which I find very confusing. Perhaps this value should also be adjustable in the configuration file so that users can play around with it and provide feedback.

guiv42:
> in page layout, scrolling occurs when X lines remain, in linear layout, scrolling occurs when X measures remain; default value X=1

I understand your point. It might be a little tricky to implement, but it makes sense. I'll have a try.

> Linear layout, 20% left margin (...)
> Perhaps this value should also be adjustable in the configuration file

Sounds good, I'll do that.
Thanks

guiv42:
Done:
- define margin parameter for continuous scrolling in configuration file
- define anticipation parameter for discrete scrolling in configuration file
- different parameters for linear/page layouts (no reason for these 2 values to be correlated, example: portrait vs. landscape display orientation)
- page layout: parameter for discrete scrolling anticipation now defines the number of *lines* to be anticipated
- (not related): can't activate/inhibit continuous scrolling while playing

@helge17 : I think it should be OK for a merge now, please tell me if you think of possible improvements/fixes

helge17:
Looks perfect now, thank you very much!

matevoux:
many thanks

Issue #986, opened 2026-01-18
Feature Request: Display Guitar Tuning in PDF Export, Print, and Print Preview


vit-herain:
Currently, the guitar tuning is not displayed in the header section during PDF export, printing, or print preview. 
This information is important for performers and readers to quickly identify the used tuning.

**Proposed Solution:**
Add the guitar tuning to the header section on the left in all exported PDFs, printed pages, and print previews. The header should include both the tuning name and the string notes, formatted as follows:

`Tuning: Dropped D (D A D G B E)`

**Benefits:**
- Provides clear tuning information for performers.
- Reduces confusion when using non-standard tunings.
- Improves the professionalism and completeness of exported/printed scores.

**Additional Notes:**
- The feature should support all tunings, displaying the correct name and string notes.
- Placement should be consistent with other header information.
- The code changes implementing this feature are almost prepared by myself; this issue serves as evidence and for tracking purposes.
- I would like to know if the maintainers and community really want and support this feature before proceeding further. 😊

guiv42:
Thanks for the proposal, that seems interesting.
A few remarks/questions:
- would it be better to display the *full* note names (i.e. with octave number): E2 A2 D3 G3 B3 E4?
- take care that some tuning may not have any name (if not corresponding to any preset)
- please avoid re-implementing a conversion from midi note pitch to note name, and re-use available methods in `app.tuxguitar.util.TGMusicKeyUtils`
- do you want me to assign this issue to you?

vit-herain:
Hi,
thanks for your points. 👍
I think full note names are not necessary here. I looked at some official transcriptions and I see there e.g. `C# F# B E G# C#`. From my perspective, it shouldn't be confused on guitar.

Yes, you can assign this to me 🙂

vit-herain:
@guiv42 PR is ready 😊 https://github.com/helge17/tuxguitar/pull/987

guiv42:
ok, thanks, I'll have a look. Please be patient and expect some delay, there are already many open topics

guiv42:
implemented by #987

PR #987, opened 2026-01-18
feat(#986): Tuning name in PDF export, print preview and print

vit-herain:
- `app.tuxguitar.song.helpers.tuning.TuningManager` is responsible for setting a correct tuning name for specific track and also for getting default tuning name (tuning with top priority in tunings.xml file)
- because `TuningManager` is used by `TGSongManager` and transitively by `TGDocumentManager` and they did not have a `TGContext` reference, I must have propagated the `TGContext` throughout constructors on many places to be always present. This makes this dependency explicit and clean and won't cause any bugs with missing context
- tuning name is set when a new song or track is created (default tuning name is applied), or a song is loaded from a file

Examples (exported PDFs):
image
image

### DIFF
--- a/common/TuxGuitar-lib/src/main/java/app/tuxguitar/graphics/control/print/TGPrintLayout.java
+++ b/common/TuxGuitar-lib/src/main/java/app/tuxguitar/graphics/control/print/TGPrintLayout.java
@@ -11,12 +11,14 @@ import app.tuxguitar.graphics.control.TGMeasureImpl;
 import app.tuxguitar.graphics.control.TGTrackImpl;
 import app.tuxguitar.graphics.control.TGTrackSpacing;
 import app.tuxguitar.song.models.TGMeasureHeader;
+import app.tuxguitar.song.models.TGString;
 import app.tuxguitar.song.models.TGTrack;
 import app.tuxguitar.ui.resource.UIFont;
 import app.tuxguitar.ui.resource.UIPainter;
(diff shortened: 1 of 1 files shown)


guiv42:
I understand the problem of data propagation with `TuningManager`and `TGContext`. But I'm really not comfortable at modifying 90 files for such a simple functional evolution.
Considering this, I have not done a detailed code review, I just made a few tests.

> tuning name is set when a new song or track is created (default tuning name is applied), or a song is loaded from a file

That does not work properly: tuning can be modified after that.
Test scenario 1:
- create new song, with all default attributes
- edit track tuning, set highest string to F4
- export pdf: it displays the old tuning *name* (inconsistent with the tuning *description*)
scenario1


Test scenario 2:
- create new song, with all default attributes
- change track tuning to D#
- export pdf

On Android app, it maintains the old tuning *name* but updates the tuning *description*
scenario2_android


Considering that tuning names are not standard, not translated, I suggest a dramatic simplification:
- only print the tuning *description*: the list of string notes
- don't display any tuning name

guiv42 (commented):
see my previous comment

vit-herain:
@guiv42 Ok, I agree, I'll simplify this according your suggestion when I have time. 👍

vitherain:
@guiv42 I implemented the minimal change you suggested.

Don't be confused - I switched my GitHub account and transfered my forked repo there. 😊

guiv42:
thanks for your contribution!

PR #990, opened 2026-01-20
Adding tempo percent controller to Main Toolbar

tdan:
- Spinner is still enabled when playing but events are ignored
- Play Mode is considered when changing percentage value. Make sure value is not greater than To value in Custom mode
- Add ChangeTempoPercentageAction


desktop/TuxGuitar/src/app/tuxguitar/app/view/toolbar/main/TGMainToolBarSectionTempo.java
```diff
@@ -56,7 +64,35 @@ public void onDispose(UIDisposeEvent event) {
...
 		});
 		this.controls.add(this.tempoLabel);
-
+		
+		this.tempoPercentSpinner = uiFactory.createSpinner(parentPanel);
+		this.tempoPercentSpinner.setVisible(true);
+		this.tempoPercentSpinner.addSelectionListener(new UISelectionListener() {
+			@Override
+			public void onSelect(UISelectionEvent event) {
+				MidiPlayer midiPlayer = MidiPlayer.getInstance(getContext());
```
guiv42: Separate the user interface from the logic.
In this class, just call actions "increment / decrement percentage", nothing more. The logic to choosing the percentage to be modified, to compute the new value, etc. shall be fully implemented in the action.
When user will call the actions to increment/decrement percentage through a keyboard shortcut, this code here will never be executed.

guiv42 (changes requested):
Main remarks:
- remove all the logic from the UI
- split the action in 2

Proposal:
- define your class `TGChangeTempoPercentageAction` as an abstract class, with a protected attribute `increment`
- implement all the logic in this class, considering `increment` can take the values +1 or -1
- define 2 classes Increment/DecrementTempoPercentageAction, extending the abstract class. In these classes constructors, call `super()` and set `this.increment` to +1 or -1
- register these 2 classes in `TGActionInstaller` and `TGActionConfigMap`, with `SHORTCUT` flag set
- add the names of these 2 actions in all translation files

please tell me if you need support. Sometimes it takes more time to try to explain than to simply do the job.

desktop/TuxGuitar/src/app/tuxguitar/app/action/impl/transport/TGChangeTempoPercentageAction.java
```diff
@@ -0,0 +1,32 @@
...
+	
+	public TGChangeTempoPercentageAction(TGContext context) {
+		super(context, NAME);
+	}
+
+	@Override
+	protected void processAction(TGActionContext context) {
+		MidiPlayer midiPlayer = MidiPlayer.getInstance(getContext());
+		MidiPlayerMode pm = midiPlayer.getMode();
+		int tempoPercent = (int) context.getAttribute(ATTRIBUTE_PERCENTAGE_VALUE);
```
guiv42: you probably get an exception here if your action is called through a keyboard shortcut

desktop/TuxGuitar/src/app/tuxguitar/app/action/installer/TGActionConfigMap.java
```diff
@@ -552,6 +553,7 @@ public void createConfigMap() {
 		this.map(TGTransportModeAction.NAME, LOCKABLE);
 		this.map(TGTransportSetLoopSHeaderAction.NAME, LOCKABLE | SHORTCUT);
 		this.map(TGTransportSetLoopEHeaderAction.NAME, LOCKABLE | SHORTCUT);
+		this.map(TGChangeTempoPercentageAction.NAME, LOCKABLE | SHORTCUT);
```
guiv42: You need also to add `DISABLE_ON_PLAY` flag, so the keyboard shortcut gets inhibited when playing

guiv42:
First review was done a bit fast (I just read the code, and did not even try to build or to test it).
Most probably the 3 different actions "change percentage", "increment percentage" and "decrement percentage" shall be not abstract, and all declared in `TGActionInstaller` and `TGActionConfigMap`. With only the last 2 ones that can be associated with a shortcut.
Just because they do not expect the same context: "change percentage" is triggered by a spinner, that will provide a numerical value. The other 2 actions do not expect any specific context value.
Still, it is healthy to implement the full logic in one single class. Best solution is probably to elaborate in "increment" and "decrement" actions a target value, store it in the context, and then call "change percentage"

desktop/TuxGuitar/src/app/tuxguitar/app/action/impl/transport/TGDecrementTempoPercentageAction.java
```diff
@@ -0,0 +1,37 @@
...
+		MidiPlayerMode mode = MidiPlayer.getInstance(getContext()).getMode();
+		int decreasedTempo;
+		
+		if (mode.getCurrentPercent() > 1) {
+			decreasedTempo = mode.getCurrentPercent() - 1;
+		} else {
+			decreasedTempo = 1;
+		}
+		
+		TGActionProcessor tgActionProcessor = new TGActionProcessor(getContext(), TGChangeTempoPercentageAction.NAME);
```
guiv42: Refer to existing implementations of actions calling other actions: re-use following pattern:
```
context.setAttribute(TGChangeTempoPercentageAction.ATTRIBUTE_PERCENTAGE_VALUE, xxx);
TGActionManager tgActionManager = TGActionManager.getInstance(getContext());
TGChangeTempoPercentageAction.execute(TGCopyAction.NAME, context);

File: desktop/TuxGuitar/src/app/tuxguitar/app/action/impl/transport/TGDecrementTempoPercentageAction.java [hunk 5]
```diff
@@ -0,0 +1,37 @@
...
+		super(context, NAME);
+	}
+
+	@Override
+	protected void processAction(TGActionContext context) {
+		MidiPlayerMode mode = MidiPlayer.getInstance(getContext()).getMode();
+		int decreasedTempo;
+		
+		if (mode.getCurrentPercent() > 1) {
+			decreasedTempo = mode.getCurrentPercent() - 1;
```
guiv42: If training mode is active, don't increment/decrement by 1, but re-use mode.customPercentIncrement

desktop/TuxGuitar/src/app/tuxguitar/app/view/toolbar/main/TGMainToolBarSectionTempo.java
```diff
@@ -108,21 +136,33 @@ public void loadProperties() {
 			// still load an icon, to make sure layout is ok (it will be refreshed after)
 			this.tempoLabel.setText("= 120");
 		}
+
+		this.tempoPercentSpinner.setMinimum(0);
```
guiv42: zero is a weird min value
See  previous remarks for min/max/increment

desktop/TuxGuitar/src/app/tuxguitar/app/view/toolbar/main/TGMainToolBarSectionTempo.java
```diff
@@ -108,21 +136,33 @@ public void loadProperties() {
...
+		this.tempoPercentSpinner.setMaximum(100);
```
guiv42: `loadProperties` is called when dialog is refreshed.
It makes no sense to set here all widget attributes that are constant (min/max, label): move these next to widgets initialization

desktop/TuxGuitar/src/app/tuxguitar/app/view/toolbar/main/TGMainToolBarSectionTempo.java
```diff
@@ -108,21 +144,33 @@ public void loadProperties() {
...
+		this.tempoPercentSpinner.setIncrement(1);
```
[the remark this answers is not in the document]
guiv42: remark still relevant

guiv42 (changes requested):

desktop/TuxGuitar/src/app/tuxguitar/app/view/toolbar/main/TGMainToolBarSectionTempo.java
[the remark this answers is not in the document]
tdan: @guiv42 please see my previous comment. Thanks :)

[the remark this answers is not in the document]
guiv42: Hi. Sorry, I cannot take the time to follow each answer to each question individually. It's a lot easier to review a PR globally 
Can you please explicitly request a review when you think it's ready?

common/TuxGuitar-lib/src/main/java/app/tuxguitar/player/base/MidiPlayerMode.java
```diff
@@ -7,6 +7,9 @@ public class MidiPlayerMode{
 	public static final int CUSTOM_DEFAULT_TEMPO_PERCENT_FROM = 80;
 	public static final int CUSTOM_DEFAULT_TEMPO_PERCENT_TO = 100;
 	public static final int CUSTOM_DEFAULT_TEMPO_PERCENT_INCREMENT = 5;
+	
+	public static final int DEFAULT_MIN_PERCENTAGE = 1;
```
guiv42, line 11: rename -> MIN_TEMPO_PERCENTAGE, MAX_TEMPO_PERCENTAGE
the term "DEFAULT" does not seem adapted.

desktop/TuxGuitar/src/app/tuxguitar/app/view/toolbar/main/TGMainToolBarSectionTempo.java
```diff
@@ -115,16 +149,18 @@ public void loadProperties() {
...
-		int tempoPercent = 100;
 
 		MidiPlayer midiPlayer = MidiPlayer.getInstance(getContext());
+		int tempoPercent = midiPlayer.getMode().getCurrentPercent();
+		
 		if ((midiPlayer.isRunning() && (midiPlayer.getCurrentTempo() != null))) {
 			tempo = midiPlayer.getCurrentTempo();
+			// update tempoPercent as midiPlayer is playing
 			tempoPercent = midiPlayer.getMode().getCurrentPercent();
-			this.tempoLabel.setIgnoreEvents(true);
```
guiv42, line 124: Why modify the implementation of `tempoLabel`?

common/TuxGuitar-editor-utils/src/main/java/app/tuxguitar/editor/action/transport/TGDecrementTempoPercentageAction.java
```diff
@@ -0,0 +1,38 @@
...
+	protected void processAction(TGActionContext context) {
+		MidiPlayerMode mode = MidiPlayer.getInstance(getContext()).getMode();
+		int decreasedTempo;
+		int decrement = 1;
+		
+		if (mode.getType() == MidiPlayerMode.TYPE_CUSTOM) {
+			decrement = mode.getCustomPercentIncrement();
+		}
+		
+		if (mode.getCurrentPercent() > decrement) {
```
guiv42, line 28: not understood why MIN value does not appear in this comparison
see my remark in `TGIncrementTempoPercentageAction`

common/TuxGuitar-editor-utils/src/main/java/app/tuxguitar/editor/action/transport/TGIncrementTempoPercentageAction.java
```diff
@@ -0,0 +1,39 @@
...
+	protected void processAction(TGActionContext context) {
+		MidiPlayerMode mode = MidiPlayer.getInstance(getContext()).getMode();
+		int increasedTempo;
+		int increment = 1;
+		
+		if (mode.getType() == MidiPlayerMode.TYPE_CUSTOM) {
+			increment = mode.getCustomPercentIncrement();
+		}
+		
+		if (mode.getCurrentPercent() <= (MidiPlayerMode.DEFAULT_MAX_PERCENTAGE - increment)) {
```
guiv42, line 28: not very readable. Proposal:
`increasedTempo = Math.min(MidiPlayerMode.DEFAULT_MAX_PERCENTAGE, mode.getCurrentPercent() + increment);`

guiv42 (commented):
still a few open points, but it's converging

PR #1034, opened 2026-03-09
Fix TGSystemOutStreamWrapper.close() to close all streams on failure

iamsanjaymalakar:
This PR improves exception safety in `TGSystemOutStreamWrapper`.

**Problem**
`TGSystemOutStreamWrapper.close()` closes multiple underlying `OutputStream`s sequentially. If one `close()` throws an `IOException`, the method exits immediately and the remaining streams are never closed, which can leave resources open.

**Fix**
Wrap each `OutputStream.close()` in a try/catch, continue closing the remaining streams even after a failure, then throw the first `IOException` and attach later ones as suppressed exceptions.

### DIFF
--- a/desktop/TuxGuitar-debug-helper/src/app/tuxguitar/debug/TGSystemOutStreamWrapper.java
+++ b/desktop/TuxGuitar-debug-helper/src/app/tuxguitar/debug/TGSystemOutStreamWrapper.java
@@ -29,8 +29,20 @@ public class TGSystemOutStreamWrapper extends OutputStream {
 
 	@Override
 	public void close() throws IOException {
-		for(OutputStream outputStream : this.outputStreams) {
-			outputStream.close();
+		IOException first = null;
+		for (OutputStream outputStream : this.outputStreams) {
+			try {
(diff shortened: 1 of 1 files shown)


guiv42:
Hi.
Can you please clarify the scenario that this PR fixes?
Said differently: can you please provide a way to test this code ?

To be honest I don't really understand the added value here. I suppose it tries to cover a degraded mode with a low probability of occurrence, in a debug-related plugin that is by default inhibited.

iamsanjaymalakar:
> Hi. Can you please clarify the scenario that this PR fixes? Said differently: can you please provide a way to test this code ?
> 
> To be honest I don't really understand the added value here. I suppose it tries to cover a degraded mode with a low probability of occurrence, in a debug-related plugin that is by default inhibited.

Thanks. This PR covers a fail-safe cleanup path for a nondeterministic failure scenario, so it is difficult to exercise with a reliable test. Even in a debug-only plugin, it is better to make cleanup best-effort, so one failing `close()` does not prevent the remaining streams from being closed.

PR #1039, opened 2026-03-13
Feature/search by keybind

### LINKED ISSUE #1018: Feature request: Search Key Bindings by Keys
Say I want to bind a key combo like Ctrl+Up. If I type Ctrl+Up in the key bindings search box, nothing comes up, so I assume it's unused. However, when I try binding it, I get a popup saying it's already in use, so I click No. Then, the only way to find what it's currently bound to is to scan through the list manually. This would be much easier if I could search the key bindings by keys.

BTW, scanning the list wouldn't be so bad except that I can't resize the window. I'll open another feature request for that.

LeonardoQF:
Resolves #1018 

Simple change that makes it so the search on the Keybind editor can find an action by both its name and the keybind it's currently assigned to.


LeonardoQF:
Ah that's fine, I'll keep it in mind for future PRs. Closing this and opening a new one to master right now.

Issue #1004, opened 2026-02-11
Feature Request  -  Copy to End feature on copy


brewern5:
Currently the upper bound auto-clamps to project length.
I’d like to propose adding a "Copy to End" button that sets the upper bound to the last measure.

If approved, I’d be happy to implement this.

guiv42:
I don't fully understand your proposal. Questions:
1. would it be useful to add as well "Copy from Start"?
2. could you please share a prototype of the concerned dialog? No need to be fully representative, a simple image drawn by hand would be sufficient, just to make sure I understand

brewern5:
I apologize that my initial proposal was not clear. If there are aspects that need clarification, please let me know.

To answer your questions:  - 

1. Yes - I believe that compliments this the copy to end feature well.
2. Attached is an image of a working prototype for the "Copy to End" feature of which I have implemented locally.

<img width="381" height="407" alt="Image" src="https://github.com/user-attachments/assets/f767c104-1ad6-4431-972e-2bfa19423996" />

If there are design guidelines or preferences - please let me know.

guiv42:
OK, thanks for the clarification. I understand that clicking on that new button will set the value of associated spinner to the last measure number. Why not.

> If there are design guidelines or preferences - please let me know.

There's no written guideline, so it's completely subjective here.
An alternative proposal:

<img width="431" height="237" alt="Image" src="https://github.com/user-attachments/assets/38a4ff16-3814-4b1c-874d-acbc7d653b15" />

(widths of widgets to be adjusted, it's just a mock-up)
Advantages:
- buttons are aligned with the spinner they are associated to (probably more intuitive)
- this example re-uses existing translated strings ("measure.first", "measure.last")

It could be interesting to disable the buttons when the associated spinner is already set to the associated value (e.g. if "From" spinner is set to 1, then button "measure.first" would do nothing)

Do you want me to assign this issue to you?

brewern5:
Yes - I would be happy to work on this issue.

I will:
- Align the first measure and last measure buttons to the spinners.  
- Disable button functionality when respective spinner is either at it's min/max state
- Size buttons to a visually balanced pattern

guiv42:
implemented by #1051

PR #1041, opened 2026-03-18
Measure Copy dialog: add First/Last range regression tests - #1004

brewern5:
Implements feature request - #1004

Adds two buttons to the Copy Measure dialog:
- First Measure: sets the From spinner to measure 1.
- Last Measure: sets the To spinner to the last measure.

Buttons are disabled when their target is already selected:

First Measure is disabled when From = 1.
Last Measure is disabled when To = last measure.

Validation:
Tested on Windows desktop.
Added and ran regression tests for range clamping and button enabled-state behavior.
Linux/macOS and Android validation still needed.


brewern5:
I apologize - this PR was not tested with the newest update, so all tests are outdated

brewern5:
Hit the wrong button.

Issue #1014, opened 2026-02-21
Feature Request: Add option to reopen previously opened files on startup


vitherain:
## Description

Forgive me if I am wrong, but currently, TuxGuitar does not automatically reopen files that were open when the application was last closed. This means users have to manually navigate and reopen their recent files each time they launch the application, which can be inconvenient, especially for those working on multiple scores or projects.

## Proposed Solution

Add a new preference setting in the desktop application to enable the automatic reopening of the last opened files upon startup. This could be implemented as a checkbox option in the Preferences dialog, under a new or existing section such as "General" or "Startup." When enabled, the application would restore the list of open tabs/files from the previous session.

The setting should:
- Be optional, with the default state set to disabled to maintain backward compatibility.
- Store the list of open files (e.g., file paths) in a configuration file or user preferences.
- Handle cases where files may no longer exist or be accessible (e.g., display a warning or skip missing files).

## Benefits

- Enhances user experience by seamlessly restoring the previous working session.
- Saves time and reduces friction for users who frequently work with multiple files.

## Additional Notes

- I can implement it myself 😊 ; this issue serves as evidence and for tracking purposes. I would like to know if the maintainers and community really want and support this feature before proceeding further. 😊

guiv42:
> Forgive me if I am wrong, but currently, TuxGuitar does not automatically reopen files that were open when the application was last closed

That's correct.

> This means users have to manually navigate and reopen their recent files each time they launch the application

You don't necessary need to manually navigate: the recently opened files are accessible in the menu File/History (recently opened files are already stored in a user config file). You can also bookmark folders using the "collection" feature (menu Tools/Collection browser).

That being said, I understand the proposal, and I think this feature can be valuable.

But warning if you want to implement it, take care of a tricky point: app may be launched by double-clicking on file(s). I guess in this case only the selected file(s) should be opened, regardless of the history. Especially, it can be a little tricky because of macOS specific behavior.
Basically, when the app is opened, it checks whether there are command line arguments specifying the file(s) to open. If there are none, a new tablature is created from a configurable template.
When you double-click on file(s) associated to TuxGuitar, Windows/Linux/FreeBSD provide the paths to TG in command-line arguments. But that's not the way macOS works.
[...]

PR #1044, opened 2026-03-22
feat(#1014): Reopen previously opened files on startup

vitherain:
## Settings option

The respective commit adds a new **"Reopen last opened files on startup"** checkbox to the General section of the Settings dialog. Introduces a new config key (`TGConfigKeys.REOPEN_LAST_FILES_ON_STARTUP`) with a default of `false` (opt-in, preserving existing behaviour). The label string is added to all 25 supported language files     
(using the English text as a placeholder in languages where no translation was provided).

image

### Core mechanism — `TGLastOpenFiles`

A new `TGLastOpenFiles` class is introduced, separate from the existing `TGFileHistory`. The distinction is intentional:

- **`TGFileHistory`** is a *recency-ordered history* used to populate the **File > History** menu. It records every file ever opened (up to a configured maximum), always placing the most recently used file first.
- **`TGLastOpenFiles`** is a *session snapshot*: it tracks exactly which files are open **right now**, and is rewritten every time documents are opened or closed. Its sole purpose is "what was open when I last quit?" — semantics that are entirely different from a history list. For example, a file can appear in
  history but not in the last-open snapshot if it was closed before quitting. Keeping these two concerns in separate classes avoids coupling the session-restore feature with the File > History behaviour.
 - TGLastOpenFiles always saves the entire list at once (save(List<URL>) clears and replaces the stored set in a single call), whereas TGFileHistory appends files one-by-one via addURL(). This difference follows directly from the semantics: TGLastOpenFiles is a snapshot of what is currently open, so it must always 
  reflect the exact current state. Whenever a document is opened or closed, TGDocumentListManager collects all currently open URLs and passes the complete list to save(). An incremental add/remove approach would require careful bookkeeping to stay consistent; replacing the whole list atomically is simpler and always
   correct.

### Startup logic — `TuxGuitar.java`

When the option is enabled, the list of last-open URLs is captured **synchronously from `TGLastOpenFiles` before `startSong()` is called**. This is critical: `startSong()` dispatches file loading asynchronously on a background thread, which eventually calls `setCurrentDocumentUri()` → `updateLastOpenFiles()`,     
overwriting `TGLastOpenFiles` with only the file being loaded. Capturing the list first and passing it explicitly to `reopenLastFiles(List<URL>)` avoids this race condition.

### macOS-specific fix — `OpenDocListener.java`

On macOS, files opened from Finder are delivered via Apple Events that SWT converts to `OpenDocument` events, not via CLI arguments. `OpenDocListener` uses a `TGActionInterceptor` to detect when the app has finished its startup phase before processing those events.

The tricky case: when "Reopen Last Files" is enabled and TuxGuitar is started without a CLI argument, `reopenLastFiles()` dispatches `TGReadURLAction` asynchronously. This action was being intercepted by `OpenDocListener`, which previously responded by calling `disconnect()` — removing both the interceptor **and**
the SWT `OpenDocument` listener. That broke Finder-triggered file opening for the remainder of the session.

The fix replaces `disconnect()` with a targeted `executeLater(removeInterceptor + processEvents)` in that branch: only the interceptor is removed, while the SWT listener stays registered for the lifetime of the application. The `executeLater` scheduling is necessary for the same reason the original `disconnect()`
used it — the interceptors list cannot be safely modified while it is currently being iterated (i.e. from inside `intercept()`).

Additionally, the `enabled` field was declared `volatile` to ensure correct visibility across threads (written on a background thread, read on the UI thread).

I tried to test various scenarios on my Mac and it worked.
When opening TuxGuitar, I used commands like `open -a "tuxguitar-9.99-SNAPSHOT-macosx-swt-cocoa.app" ~/Downloads/degug_song.tg` or `open -a "tuxguitar-9.99-SNAPSHOT-macosx-swt-cocoa.app" --arg ~/Downloads/degug_song.tg`. I also tried to open specific files via Finder (i drag & dropped files to the app icon). One scenario was to open file that was already present on last open files list and it opened it only once.


desktop/TuxGuitar/src/app/tuxguitar/app/TuxGuitar.java
```diff
@@ -178,6 +199,42 @@ private void startDefaultSong() {
 		tgActionProcessor.process();
 	}
 
+	private void reopenLastFiles(List<URL> urls) {
+		List<URL> notFoundUrls = new ArrayList<URL>();
```
vitherain: I wanted to avoid an error message that the file was not found (renamed or moved file). Instead, I wanted to remove it right away from the list and not try to open it at all.

common/resources/lang/messages_bg.properties
```diff
@@ -573,6 +572,7 @@ settings.config.main.options=Настройки
 settings.config.main.table.auto-size.enabled=Включване автоматично оразмеряване на таблица
 settings.config.main.splash-enabled=Показване на лого при стартиране
 # settings.config.main.play-when-moving=
+settings.config.main.reopen-last-files=Повторно отваряне на последните файлове при стартиране
```
guiv42: Where do all these translations come from?
I can see 2 options:
- option 1: retrieved from another open-source software
- option 2: automated translation with AI

If this is option 2, then please remove all automated translations, we keep this task for human contributors.

desktop/TuxGuitar/src/app/tuxguitar/app/TuxGuitar.java
```diff
@@ -150,8 +155,24 @@ private void startUIContext(URL url) {
...
 
+		// Capture the last-open list NOW, before startSong() dispatches async
+		// file loading that will call setCurrentDocumentUri() →
+		// updateLastOpenFiles(), overwriting TGLastOpenFiles with only the
+		// Java-arg file before reopenLastFiles() has a chance to read it.
+		final List<URL> lastOpenUrls = getConfig().getBooleanValue(TGConfigKeys.REOPEN_LAST_FILES_ON_STARTUP)
+				? new ArrayList<>(TGLastOpenFiles.getInstance(this.context).getURLs())
+				: new ArrayList<>();
+
 		this.startSong(url);
```
guiv42: Can there be some concurrence between the thread reopening last open files and the one created for `startSong` ?
Which song will have the focus once everything is correctly opened? Is it deterministic or does it depend from the time needed to open each song?

What do you think of the following approach?
- set `lastOpenUrls` as in line 162
- if `url` is valid, then append `url` to `lastOpenUrls`
- delete `this.startSong(url)`: only rely on `reopenLastFiles` (that may need to be renamed)

desktop/TuxGuitar/src/app/tuxguitar/app/TuxGuitar.java
```diff
@@ -178,6 +199,42 @@ private void startDefaultSong() {
...
+		for(TGDocument doc : TGDocumentListManager.getInstance(this.context).getDocuments()) {
+			if(doc.getUri() != null) {
+				try {
+					currentUrls.add(doc.getUri().toURL());
+				} catch (Exception e) {
+					// skip invalid URIs
+				}
+			}
+		}
+		for(URL url : lastOpenUrls) {
```
guiv42, line 204: That would be better to iterate in this list in the reverse order.
Currently, when last session is restored, the order of the tabs is inverted (compared to the order when the app was closed).

desktop/TuxGuitar/src/app/tuxguitar/app/TuxGuitar.java
```diff
@@ -150,8 +155,24 @@ private void startUIContext(URL url) {
...
+		// Java-arg file before reopenLastFiles() has a chance to read it.
+		final List<URL> lastOpenUrls = getConfig().getBooleanValue(TGConfigKeys.REOPEN_LAST_FILES_ON_STARTUP)
+				? new ArrayList<>(TGLastOpenFiles.getInstance(this.context).getURLs())
+				: new ArrayList<>();
+
 		this.startSong(url);
 		this.setInitialized(true);
+
+		if( !lastOpenUrls.isEmpty() ) {
+			TGSynchronizer.getInstance(this.context).executeLater(new Runnable() {
```
guiv42, line 178: Why `executeLater`?
If I understand correctly, `reopenLastFiles` relies on `TGActionProcessor` instances to open the files (one instance per file). And as `.process()` is called without an argument, each file opening action will be processed asynchronously in a new thread.

guiv42 (commented):
Thanks for your contribution, and sorry for the long delay for review. Especially, thanks for taking the time to write a precise PR description, that really helps the review process.


Just one more question: how complex would it be to restore the focus where it was when app was closed? Two cases would need to be considered:
- app called without any argument: that would be useful to restore the focus (if possible)
- app called with at least one argument: set the focus to the last song in these arguments (if it exists, else focus can be set arbitrarily).

That's just a suggestion ("nice to have", definitely not mandatory).

(note: I rely on your tests for the macOS configuration)

common/resources/lang/messages_bg.properties
vitherain: I removed all translations besides EN and CS that are 100 % correct. 😉

desktop/TuxGuitar/src/app/tuxguitar/app/TuxGuitar.java
vitherain: I tested this and the tab order is actually preserved correctly. 🤔 

Newly opened file is appended to the end of the list, so iterating `lastOpenUrls` in order produces the same tab order as when the app was closed. Reversing the iteration would in fact invert the order.

desktop/TuxGuitar/src/app/tuxguitar/app/TuxGuitar.java
vitherain: Actually, `executeLater` turns out to be necessary. Without it, `reopenLastFiles` fires `TGReadURLAction` almost simultaneously with `startDefaultSong`'s `TGLoadTemplateAction`. This creates a race in `OpenDocListener`: if `TGReadURLAction` is intercepted first, the interceptor is not yet removed when `TGLoadTemplateAction` arrives, so both interception branches run. The `TGLoadTemplateAction` branch spawns a background thread that calls `processEvents()` holding `synchronized(this.getClass())` while invoking `getPlayer().reset()`. If a Finder-triggered `OpenDocument` event arrives at that moment, `handleEvent()` on the UI thread tries to acquire the same lock — deadlock, app freezes.   

`executeLater` defers `reopenLastFiles` to the next event-loop iteration, giving `TGLoadTemplateAction` enough time to be intercepted first and remove the interceptor. By the time `reopenLastFiles` fires its `TGReadURLAction`, the interceptor is gone and the dual-interception race cannot occur.

desktop/TuxGuitar/src/app/tuxguitar/app/TuxGuitar.java
vitherain: I have implemented your suggested approach: `startSong` is removed, the initial document is marked as unwanted upfront, and both the session files and the CLI argument (if valid and the file exists) are collected into a single list passed to `reopenLastFiles`. 

If the list is empty, `startDefaultSong` is called as a fallback. This eliminates the concurrence between the two separate loading paths. Focus remains non-deterministic when multiple files are opened simultaneously (each on its own background thread), but that is a pre-existing limitation.

vitherain:
> Thanks for your contribution, and sorry for the long delay for review. Especially, thanks for taking the time to write a precise PR description, that really helps the review process.
> 
> Just one more question: how complex would it be to restore the focus where it was when app was closed? Two cases would need to be considered:
> 
> * app called without any argument: that would be useful to restore the focus (if possible)
> * app called with at least one argument: set the focus to the last song in these arguments (if it exists, else focus can be set arbitrarily).
> 
> That's just a suggestion ("nice to have", definitely not mandatory).
> 
> (note: I rely on your tests for the macOS configuration)

Thank you for the kind words and for the thorough review!

Regarding focus: currently, when multiple files are opened on startup, focus is non-deterministic — it lands on whichever background thread completes last. Properly restoring it would require saving the active document's URL at session end and then waiting for all files to finish loading before switching to it. I'd rather keep this PR focused on the core reopening feature and track focus restoration as a separate issue, if that works for you.

desktop/TuxGuitar/src/app/tuxguitar/app/TuxGuitar.java
guiv42: Wrong interpretation on my side.
Since all files are opened asynchronously, the order of the tabs at the end of the process is not really deterministic. And it's independent from the order of the iteration loop

desktop/TuxGuitar/src/app/tuxguitar/app/TuxGuitar.java
guiv42: Sorry, I still don't understand.
> Actually, `executeLater` turns out to be necessary. Without it, `reopenLastFiles` fires `TGReadURLAction` almost simultaneously with `startDefaultSong`'s `TGLoadTemplateAction`.

I understand that calls to `reopenLastFiles` and `startDefaultSong` are exclusive, so how could they compete?

guiv42 (commented):
Thanks for your answers and the PR update. Still a few open points.

I just detected one critical issue (was introduced by 32e34f7fc73cea729c903f1a06753edad0624bec). Test scenario:
- create and open 3 .tg files
- check option to reopen files at startup
- close app
- reopen app

The 3 songs are correctly re-opened, but it seems there is a synchronization issue somewhere (the kind of problem I really dislike). I get several occurrences of the following exception:
```
java.lang.NullPointerException: Cannot invoke "media.sound.AudioSynthesizer.loadInstrument(javax.sound.midi.Instrument)" because "this.synth" is null
        at app.tuxguitar.midi.synth.impl.GervillProcessor.loadInstrument(GervillProcessor.java:113)
        at app.tuxguitar.midi.synth.impl.GervillProcessor$1.onCreate(GervillProcessor.java:104)
        at app.tuxguitar.midi.synth.impl.GervillSoundbankFactory.createInstrument(GervillSoundbankFactory.java:46)
        at app.tuxguitar.midi.synth.impl.GervillSoundbankFactory.access$000(GervillSoundbankFactory.java:18)
        at app.tuxguitar.midi.synth.impl.GervillSoundbankFactory$1.run(GervillSoundbankFactory.java:33)
        at java.base/java.lang.Thread.run(Thread.java:1474)
```

I don't get any exception if I uncheck the option to reopen files, even if I call TG with a path to a valid file as argument

vitherain:
@guiv42 Sorry, I am a bit busy these days. I hope I'll look at this again as soon as possible. Thanks for your patience.

guiv42:
No problem at all. One important thing in this project: you can consider nothing is urgent.

I did not investigate at all this issue, but I was just thinking about a possible simpler alternative. Considering that your current branch:
- builds and maintains consistently an ordered list of previously opened files
- adds in this list the eventual files passed with command line arguments

Wouldn't it be simpler to open *sequentially* all files in this list, from one single thread? That would make the order of the tabs once everything is finished deterministic. Could it eventually solve the exception I mentioned? (no idea...)

guiv42:
@vitherain: could you please have a look at my [dev branch](https://github.com/guiv42/tuxguitar/tree/reopen-files)?

I started this branch from the current state of this PR. Then I made sure all songs are opened from one single thread when app starts, sequentially. Advantages:
- the order of songs after startup is deterministic (it's preserved)
- it reduces the number of threads launched in parallel during startup (safer)
- the logic is closer to the current implementation, to minimize the risk

Note that putting everything in the same thread did not fix the exception raised by Gervill. I had to remove 1 line (that I did not really understand) to fix it.

One important open point: I have not tested at all in macOS environment.
Since the logic is now very close to the current implementation, I would bet that your modification of the macOS plugin should be reverted.

Waiting for your feedback.

guiv42:
I just understood the Gervill exception, my current implementation in my dev branch is incorrect.
Basically, in Gervill configuration:
- when one song is opened, instruments are initialized asynchronously
- when changing selected song, instruments configuration is reset

When we're loading several songs in a loop, these 2 operations are executed quite fast, and conflict. Fix in progress.

guiv42:
I have deleted the first dev branch, and created [another one](https://github.com/guiv42/tuxguitar/tree/reopen-files2) with a clean fix of the Gervill issue.

Again I started from current state of this PR. See the commits, chronologically:
1. regroup all songs opening in the same thread. Restore the initial logic of app startup.
2. fix properly the Gervill exception
3. revert the macOS plugin modification

@vitherain : could you please have a look at it?
Especially, item 3 above is only based on a theoretical analysis, because I cannot test in macOS environment.

Edit: a small modification is still required in Gervill for a complete fix (addition of a `synchronized` condition). I'll do that later.

guiv42:
`synchronized` flag added in my dev branch

Issue #921, opened 2025-10-14
Option to discard all changes on exit


bergentroll:
There is a usual case (as for me):
 
1. Start TuxGuitar for training 
2. Open a project
3. Disable a track the user would like to live play
4. Play with TuxGuitar
5. Open another project
6. GOTO 2 until the user is tied
...
8. Close TuxGuitar

On p. 7 (close) TuxGuitar asks to save changes per every track which may become annoying. Is it a chance to implement an "Exit now" dialog option to discard changes for remain tabs?

LeonardoQF:
Could you specify which tuxguitar version and OS you are on? I could work on the issue given those details.

guiv42:
> Could you specify which tuxguitar version and OS you are on?

I don't understand why you ask this question. Current behavior and implementation are completely OS-agnostic in this case.

LeonardoQF:
> > Could you specify which tuxguitar version and OS you are on?
> 
> I don't understand why you ask this question. Current behavior and implementation are completely OS-agnostic in this case.

I haven't been able to reproduce the behavior described in the issue. Even when following the exact steps, the only dialog box that shows up when closing the app is the regular "Song has unsaved changes, would you like to save them first?"; which led me to believe maybe it was a version-specific dialog box.

guiv42:
OK, understood.
Here's what I get with TG 2.0.1 on Linux (openSuse tumbleweed). I would be surprised if the behavior differs on other OSes (though I cannot test)

https://github.com/user-attachments/assets/6b35830a-a966-48fd-9a95-e204876d8cf1

LeonardoQF:
Ah I see now. The way it was phrased, I thought a confirm dialog would appear for each modified track, instead of each modified song (having modified tracks). Now i get it. 
I do wonder how it'd be best to implement the extra "discard all" UX/UI-wise. Maybe add it to the right of cancel?

guiv42:
I don't really know, I'm not very comfortable with user interface definition.
The best should probably be to get some inspiration from other implementations of well-known software. Just as an example, here is what it looks like with Gimp:

<img width="437" height="307" alt="Image" src="https://github.com/user-attachments/assets/889bb429-967d-44cc-b69a-5d098d96656d" />

helge17:
Two remarks on PR #1049:

[...]

PR #1049, opened 2026-03-26
Add Discard All option when closing with multiple unsaved documents

alexdomini:
Should fix issue #921.

When multiple songs have unsaved changes and the user closes TuxGuitar, a **Discard All** button now appears in the save-changes dialog. Clicking it discards all remaining unsaved changes at once, without prompting for each document individually.

The button only appears when there is more than one unsaved document remaining, so the single-document case is unchanged.

Also updated `tuxguitar.bat` to fall back to system `java` when no local `jre` folder is present, making it easier to run from source without bundling a JRE.

### DIFF
--- a/desktop/build-scripts/common-resources/common-windows/tuxguitar.bat
+++ b/desktop/build-scripts/common-resources/common-windows/tuxguitar.bat
@@ -5,7 +5,7 @@ SET "TG_START=%cd%"
 cd %~dp0
 SET "TG_DIR=."
 ::JAVA
-SET "JAVA=jre\bin\java"
+IF EXIST "jre\bin\java.exe" (SET "JAVA=jre\bin\java") ELSE (SET "JAVA=java")
 ::LIBRARY_PATH
 SET "JAVA_LIBRARY_PATH=%JAVA_LIBRARY_PATH%;lib\"
 ::CLASSPATH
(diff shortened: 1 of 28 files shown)


desktop/build-scripts/common-resources/common-windows/tuxguitar.bat
```diff
@@ -5,7 +5,7 @@ SET "TG_START=%cd%"
 cd %~dp0
 SET "TG_DIR=."
 ::JAVA
-SET "JAVA=jre\bin\java"
```
guiv42: Zero knowledge on windows on my side.
@helge17 is that OK?

@alexdomini : please next time submit separate PRs for different features. This modification is absolutely unrelated to the PR title, and not even mentioned in the commit message.

guiv42 (commented):
Test scenario:
- 2 songs modified and unsaved opened
- 1 song open without any modification
- Focus is on the unmodified song
- close the app

The displayed message "**The song** has unsaved changes" is not perfectly adapted: "One song ..." would be more appropriate (that's a detail).

Another general remark: still with the same test scenario, focus changes automatically to the first unsaved song when user tries to close app. If user clicks "Cancel", focus remains on this unsaved song.
I don't really know if it's a problem (and I don't have any idea how to avoid this focus change).
What do you think?

alexdomini:
- Reverted the unrelated tuxguitar.bat change (will submit in a separate PR)
 - Added #file.discard-all-changes= and #file.save-changes-question-many= placeholders to all 25 language files
 - Added a separate message key file.save-changes-question-many used when multiple songs have unsaved changes ("One song has unsaved changes...")
 - Regarding the focus change on Cancel: I agree it's not ideal, but fixing it would require a more significant refactor of the interceptor flow. Happy to address it in a separate issue if you think it's worth it.

guiv42:
Thanks!

PR #1066, opened 2026-04-18
Fix issue 865: panning and mixing not power conservative

ycollet:
Fix some issues with the mixing:
- the panning law is not power conservative: when a track is centered, the level of the track is reduced by -3db. I used cos / sin to compute the balance
- the mixing is not power conservative: the more you add track, the quieter the main signal become. I divided the sum of tracks by sqrt(N) instead of N.

### DIFF
--- a/desktop/TuxGuitar-synth/src/app/tuxguitar/midi/synth/TGAudioBuffer.java
+++ b/desktop/TuxGuitar-synth/src/app/tuxguitar/midi/synth/TGAudioBuffer.java
@@ -111,13 +111,20 @@ public class TGAudioBuffer {
 			short maxValue = Short.MAX_VALUE;
 			short minValue = Short.MIN_VALUE;
 
+			// Constant-power mix: mix() stores the running average (sum/N).
+			// Multiplying back by sqrt(N) gives sum/sqrt(N), which keeps the
+			// total output power constant as instruments are added regardless
+			// of N (for uncorrelated sources).  The previous factor of N
+			// reconstructed the raw sum, whose amplitude grew linearly with N.
+			float sqrtN = (float) Math.sqrt(this.buffers);
+
 			for (int i = 0; i < ( BUFFER_SIZE / 2 ); i++) {
 				for (int j = 0; j < CHANNELS; j++) {
 					sampleValue = 0;
 					for( int n = 0; n < 2; n ++){
 						sampleValue |= ((this.buffer[(index + n)] & 0xff) << ( 8 * (BIGENDIAN ? 1 - n : n)));
 					}
-					ampValue = (short) (Math.max(Math.min(sampleValue * this.buffers, maxValue), minValue) & 0xFFFF);
+					ampValue = (short) (Math.max(Math.min(Math.round(sampleValue * sqrtN), (int) maxValue), (int) minValue) & 0xFFFF);
 
 					this.buffer[index++] = (byte) ((BIGENDIAN ? ((ampValue & 0xFF00) >> 8) : ((ampValue & 0x00FF) >> 0) ) & 0xFF);
 					this.buffer[index++] = (byte) ((BIGENDIAN ? ((ampValue & 0x00FF) >> 0) : ((ampValue & 0xFF00) >> 8) ) & 0xFF);
--- a/desktop/TuxGuitar-synth/src/app/tuxguitar/midi/synth/TGSynthChannelProcessor.java
+++ b/desktop/TuxGuitar-synth/src/app/tuxguitar/midi/synth/TGSynthChannelProcessor.java
@@ -217,11 +217,17 @@ public class TGSynthChannelProcessor {
 
 	public void balance(TGAudioBuffer buffer, int volume, int balance) {
 		synchronized (this.lock) {
+			// Constant-power panning law: use sine/cosine so that the total
+			// power (left² + right²) remains constant across all pan positions.
+			// At center (balance=63.5) both channels get cos(π/4) ≈ 0.707,
+			// whereas a linear law would give 0.5 — a 3 dB centre-hole.
+			float pan = (balance / 127f) * ((float) Math.PI / 2f);
+			float gain = volume / 127f;
 			buffer.balance(new float[] {
 				// left channel
-				(volume / 127f) * ((127 - balance) / 127f),
+				gain * (float) Math.cos(pan),
 				// right channel
-				(volume / 127f) * ((balance) / 127f)
+				gain * (float) Math.sin(pan)
 			});
 		}
 	}


guiv42:
Thanks for the contribution, I've been turning around this issue for some time already, yet without success.

Just a question: any idea about a possible "rigorous" way to test this? It's always possible to test different configurations and try to compare the perceived volume by ear, but this is not very objective.

I just tried the following test:
1. configure TuxGuitar synthesizer in sound settings
2. start from an empty song
3. Copy-paste from a drums track of an existing song
4. play the song, and listen
5. click 15 times on "add track" button, play the song and listen
6. delete the 15 added tracks, play the song and listen

With both master branch and your branch, volume at step 5 is significantly lower than at step 4. I would say that your branch sounds better, but I did not find a method to perform a really reliable comparison.

What is more surprising to me: perceived volumes are very similar (identical?) at steps 5 and 6 (only tested with your branch). It seems that deleting tracks leaves number of channels in song unchanged: unused ones are probably not cleared, I'll have a look at this.

ycollet:
Thanks a lot for your tests. I will do some more on my side and try and keep you informed.

ycollet:
I made a test with the current version of tuxguitar.
I created a first tg file with 1 track: a Saw synth track with a long note and a balance completely on the left.
I recorded the sound via audacity.
I created a second tg file with 10 tracks: 10 tracks with the saw synth and only a long note on the first track, the balance on the left.
I recorded the sound via audacity.
I measured the RMS value between 2 lows of a cycle: -21.3db in both cases ...
image
So, I started to wonder if there was a "master" level parameter in the guitarpro format which was not read by tuxguitar but still played ...
I am a little bit puzzled by this result because normally, when in the mixer you divide by N (the number of tracks), then there is a global level problem (the more you add tracks the quieter the master become).

Need to investigate ...

ycollet:
Tried to generate the tg files using 2 differents methods:
The first one had 10 tracks but with the same instrument ...
- new 10 saw synth tracks but with saw synth as different instruments. A not in the first track and nothing in the other tracks. RMS value -> -21.3 db
- new 10 saw synth tracks with 10 saw synth instruments. A note for all 10 tracks. I set the volume at the max for the first track and I set the volume to 0 for the 9 others. RMS value -> -21.3dB

A note: I am using the fluidsynth plugin ... Maybe the mixing is done correctly in the fluidsynth plugin.

ycollet:
I used Claude to try to find out ...
```
⏺ Let me trace through the math carefully to analyze your test results and the correctness of the patch.                                                                                                                                   
                                                    
  What the code actually does                                                                                                                                                                                                              
                                                                                                                                                                                                                                           
  mix() builds a running arithmetic mean:                                                                                                                                                                                                  
  - After N tracks: stored value = (s₁ + s₂ + ... + sN) / N                                                                                                                                                                                
                                                                                                                                                                                                                                           
  Original clip(): multiplies by N → output = sum(s₁..sN)                                                                                                                                                                                  
                                                                                                                                                                                                                                           
  Patched clip(): multiplies by sqrt(N) → output = sum / sqrt(N)                                                                                                                                                                           
                                                                                                                                                                                                                                           
  Why your test shows -21.3 dB for both cases on the official release                                                                                                                                                                      
   
  With 10 tracks, only track 1 active (signal s, others = 0):                                                                                                                                                                              
                                                                                                                                                                                                                                         
  - mix stores: (s + 0 + ... + 0) / 10 = s/10                                                                                                                                                                                              
  - original clip: s/10 * 10 = s → same level as 1 track ✓                                                                                                                                                                               
                                                                                                                                                                                                                                           
  The original multiply-by-N correctly reconstructs the sum. Since the sum of 1 signal + 9 zeros = that signal, the RMS is unchanged. Your test confirms the original code behaves correctly for this case.                                
                                                                                                                                                                                                                                           
  What the patch actually changes                                                                                                                                                                                                          
                                                                                                                                                                                                                                         
  If you ran the same test on the patched version:                                                                                                                                                                                         
  - clip: s/10 * sqrt(10) = s/sqrt(10) → about -10 dB lower than 1-track case
                                                                                                                                                                                                                                           
  That's a regression: a solo instrument sounds quieter just because the score has more (silent) tracks.                                                                                                                                 
                                                                                                                                                                                                                                           
  The test that actually reveals the difference                                                                                                                                                                                          
                                                                                                                                                                                                                                           
  Your test design (1 active, N-1 silent) doesn't stress the mixing law. The meaningful test is:                                                                                                                                           
   
  ┌───────────────────────────────┬───────────────┬────────────────┐                                                                                                                                                                       
  │           Scenario            │ Original (×N) │  Patch (×√N)   │                                                                                                                                                                     
  ├───────────────────────────────┼───────────────┼────────────────┤
  │ 1 track, signal A             │ A             │ A              │
  ├───────────────────────────────┼───────────────┼────────────────┤
  │ 2 identical tracks, signal A  │ 2A (+6 dB)    │ √2·A (+3 dB)   │                                                                                                                                                                       
  ├───────────────────────────────┼───────────────┼────────────────┤                                                                                                                                                                       
  │ 10 identical tracks, signal A │ 10A (clips)   │ √10·A (+10 dB) │                                                                                                                                                                       
  └───────────────────────────────┴───────────────┴────────────────┘                                                                                                                                                                       
                                                                                                                                                                                                                                         
  For a music player, the original behavior (sum) is actually correct: two instruments playing together should be louder than one. The patch implements "constant-power" mixing, which is appropriate for crossfading/parallel processing, 
  but not for multi-instrument playback.                                                                                                                                                                                                 
                                                                                                                                                                                                                                           
  Conclusion                                                                                                                                                                                                                               
   
  The panning fix is correct (sine/cosine law genuinely fixes a -3 dB center hole). The mixing fix is questionable:                                                                                                                        
                                                                                                                                                                                                                                         
  - The original code computed the sum — appropriate for a music player                                                                                                                                                                    
  - The patch computes sum/sqrt(N) — makes active tracks quieter based on total track count, which is wrong                                                                                                                              
  - Your Audacity measurement gives no evidence of a problem in the original mixing behavior                                                                                                                                               
  - The "power conservation problem" you perceived in some scores is worth investigating with the correct test: compare 1 track playing vs. 2 of the same tracks playing simultaneously                                                    
                                                                                                                                                                                                                                         
❯ For the perception problem, I think it is maybe due to some kind of master level parameter used by GuitarPro, read by TuxGuitar, but not tunable in TuxGuitar. So, for some scores, I feel them       
  quieter because of this hidden parameter ...                                                                                                                                                                                             
                                                                                                                                                                                                                                           
⏺ That's a plausible hypothesis. Guitar Pro files (especially .gp5/.gpx) have a master volume/RSE master level field that gets read during import. TuxGuitar may apply it during MIDI rendering but not surface it in the mixer UI, leaving
   some scores permanently quieter with no obvious way to compensate.                                                                                                                                                                      

  If you ever want to investigate, the import code for GP file master volume would be in the GP4/GP5 readers — something like TGGPFileImporter or similar. Searching for masterVolume or masterEffect in the importer would be a good      
  starting point if the hypothesis feels worth confirming.   
```

So, I will revert the sqrt part and leave only the fix on the panning :)

guiv42:
Don't rush to revert anything, I need to spend more time to better understand.

If I understand correctly, your code only modifies the TuxGuitar synthesizer. So if you are using FluidSynth plugin it seems normal to me that you don't see any difference.
That's the tricky part for the testing: it's OK to record the output in several configurations for comparison, but we need to make sure the sound generation follows the correct path in this test. Leave me some time to dig in...

guiv42:
I could finally find some time to perform a few tests, and a comparison between 2.0.1 and this PR. For each scenario, I recorded the output with Audacity and measured the RMS value.

Tests scenarios:
- scenario 0 (ref): from default template, add a single note with an instrument with constant volume (flute). Keep the drumkit channel present, but empty
- scenario 1: from reference scenario, delete the unused drumkit channel
- scenario 2: from reference scenario, add 8 empty tracks (menu Track/Add track), this adds as many unused channels
- scenario 3: from reference scenario, clone the first track 4 times. This means 5 tracks performing the very same (correlated) sound
- scenario 4 : from reference scenario, add 4 tracks with the same note, but all different instruments (again instruments with a constant volume, like organ, strings, ...)

Here are the results (RMS values, dB)
| Scenario | TG 2.0.1 | this PR |
| ------------- | ------------- | ------------- |
| reference | -27.5 | -27.6 |
| scenario 1 | -27.5 | -24.6 |
| scenario 2 | -27.7 | -34.7 |
| scenario 3 | -14 | -14 |
| scenario 4 | -23.9 | -27.6 |

I don't explain everything, but from an empirical point of view it seems the original version (2.0.1) performs better.
@ycollet: I think your suggestion to revert the sqrt part and only keep the panning modification seems OK.

ycollet:
Thanks a lot for these tests. I will update the PR and remove the mix part. I will just leave the balance part ...

ycollet:
PR updated

guiv42 (approved):
OK

ycollet:
Thanks a lot for the merge !

PR #1067, opened 2026-04-20
Update INSTALL.md: Add missing build step for Java core modules on Ge…

vi-volatus:
…neric GNU/Linux

Without running mvn install -DskipTests in the desktop folder first, the build fails with: Could not find artifact app.tuxguitar:tuxguitar-synth-lv2:jar:9.99-SNAPSHOT Tested on CachyOS.


guiv42:
I cannot see any link between lv2 plugin and unit tests. There must be something specific to your configuration.

I'll add a section "troubleshooting" to the INSTALL.md to describe the possibility to skip unit tests. But I don't think it's a good idea to add a step that should not always be needed

PR #1080, opened 2026-05-26
Add NSM support

ycollet:
Add NSM (Non Session Manager) support

This PR adds a TuxGuitar plugin that integrates with NSM-compatible session managers such as RaySession, New Session Manager, and Agordejo. When $NSM_URL is set in the environment, TuxGuitar announces itself to the session manager and participates fully in the session lifecycle.

---
What was added

New plugin: desktop/TuxGuitar-nsm/

- NSMPlugin.java — implements TGEarlyInitPlugin so the session directory is known before any config file is read. Blocks briefly (≤ 5 s) for /nsm/client/open, then redirects all config I/O to the session directory.
- NSMClient.java — OSC/UDP client. Handles the full NSM protocol:
  - /nsm/server/announce at startup
  - /nsm/client/open — loads the session song file or an empty template for new sessions
  - /nsm/client/save — saves via TGWriteFileAction, sends reply when complete
  - /nsm/client/quit — routes through TGExitAction (same path as File › Exit, shows "save before exit?" dialog if there are unsaved changes)
  - TGActionInterceptor — suppresses the default empty-song load at startup so the session file is opened instead
- NSMConfigPropertiesHandler.java — redirects "config" resource reads/writes to the session directory, keeping each session's settings isolated.
- NSMSignal.java + jni/app_tuxguitar_nsm_NSMSignal.c — JNI bridge that installs a C-level sigaction(SIGTERM) handler via a self-pipe trick, as a fallback for direct kill -TERM sent to the JVM.
- osc/OSCMessage.java — minimal OSC 1.0 encoder/decoder (no external dependency).

Session directory layout

Each session gets a self-contained directory provided by the session manager:
{sessionDir}/tuxguitar.tg                    ← song file
{sessionDir}/tuxguitar.cfg                   ← main app config
{sessionDir}/plugins/tuxguitar-*.cfg         ← per-plugin configs
{sessionDir}/tuxguitar-plugin-settings.cfg   ← plugin enabled/disabled state

SIGTERM handling (RaySession compatibility)

Session managers such as RaySession terminate clients with SIGTERM rather than /nsm/client/quit. Intercepting SIGTERM from Java is non-trivial because SWT/GTK and libjack both override the JVM's sigaction(SIGTERM) handler — libjack in particular reinstalls its handler each time a JACK client connects or disconnects, which can happen at the very moment the session is closing.

The solution uses the bash launch script as an intermediary:

- tuxguitar.sh exports NSM_LAUNCHER_PID=$$ and runs Java in the background. A trap converts SIGTERM → SIGUSR2 and forwards it to the Java child.
- NSMClient announces NSM_LAUNCHER_PID to the session manager, so SIGTERM is sent to the bash wrapper rather than the JVM.
- A sun.misc.Signal("USR2") handler in NSMClient receives SIGUSR2 and calls handleQuit(). JACK never installs a SIGUSR2 handler, so this fires reliably every time.
- A JNI-based 100 ms watchdog continuously reinstalls the C sigaction as a fallback for direct kill -TERM <java_pid>.

Build integration

- tuxguitar-linux-swt/pom.xml, tuxguitar-linux-jfx/pom.xml, tuxguitar-linux-qt/pom.xml — NSM jar included in each Linux distribution; the native-modules profile of the SWT build compiles libtuxguitar-nsm-jni.so inline.
- desktop/pom.xml — dependency management entry.
- build-scripts/native-modules/tuxguitar-nsm-linux/pom.xml — standalone Maven module for the JNI native library (not used by the SWT reactor to avoid duplicate-module conflicts, but available for standalone builds).

---

I tested this on Fedora 43 and the TuxGuitar configuration files were stored in the NSM session directory (here, I used Ray Session):
```
$ ls -lR TuxGuitar.tuxguitar/
TuxGuitar.tuxguitar/:
total 28
drwxr-xr-x. 2 collette collette  4096 13 mai   18:12 plugins
-rw-r--r--. 1 collette collette  3619 21 mai   18:04 tuxguitar.cfg
-rw-r--r--. 1 collette collette 19044 21 mai   18:04 tuxguitar.tg

TuxGuitar.tuxguitar/plugins:
total 16
-rw-r--r--. 1 collette collette 329 13 mai   18:13 tuxguitar-fluidsynth.cfg
-rw-r--r--. 1 collette collette  66 13 mai   18:12 tuxguitar-gtp.cfg
-rw-r--r--. 1 collette collette  47 13 mai   18:12 tuxguitar-jack.cfg
-rw-r--r--. 1 collette collette 123 13 mai   18:12 tuxguitar-jsa.cfg
```

Claude added some JNI code to be able to register a SIGTERM handler. Several times, the SIGTERM handler was not working because Jack seems to intercept the SIGTERM signal and not rethrowing it.
Now, with these last changes, when I close a session in Ray Session, TuxGuitar closes gracefully.

### DIFF
--- a/android/TuxGuitar-android/src/app/tuxguitar/android/view/dialog/repeat/TGRepeatCloseDialog.java
+++ b/android/TuxGuitar-android/src/app/tuxguitar/android/view/dialog/repeat/TGRepeatCloseDialog.java
@@ -46,26 +46,29 @@ public class TGRepeatCloseDialog extends TGModalFragment {
 		if( repeatCloseDefault < 1 ) {
 			repeatCloseDefault = 1;
 		}
+		// Show total play count (stored value + 1).
+		int displayDefault = repeatCloseDefault + 1;
 
 		ArrayAdapter<Integer> arrayAdapter = new ArrayAdapter<Integer>(getActivity(), android.R.layout.simple_spinner_item, createRepeatValues());
 		arrayAdapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item);
 
 		Spinner spinner = (Spinner) this.getView().findViewById(R.id.repeat_close_dlg_count_value);
 		spinner.setAdapter(arrayAdapter);
-		spinner.setSelection(arrayAdapter.getPosition(Integer.valueOf(repeatCloseDefault)));
+		spinner.setSelection(arrayAdapter.getPosition(Integer.valueOf(displayDefault)));
 	}
 
 	public Integer[] createRepeatValues() {
-		Integer[] items = new Integer[101];
+		// Values 2..101 represent "play N times" (stored as N-1 internally).
+		Integer[] items = new Integer[100];
 		for (int i = 0; i < items.length; i++) {
-			items[i] = Integer.valueOf(i);
+			items[i] = Integer.valueOf(i + 2);
 		}
 		return items;
 	}
 
 	public int parseRepeatValue() {
 		Spinner spinner = (Spinner) this.getView().findViewById(R.id.repeat_close_dlg_count_value);
-		return ((Integer)spinner.getSelectedItem()).intValue();
+		return ((Integer)spinner.getSelectedItem()).intValue() - 1;
 	}
 
 	public void changeRepeatClose() {
--- a/common/TuxGuitar-lib/src/main/java/app/tuxguitar/graphics/control/TGMeasureImpl.java
+++ b/common/TuxGuitar-lib/src/main/java/app/tuxguitar/graphics/control/TGMeasureImpl.java
@@ -932,7 +932,7 @@ public class TGMeasureImpl extends TGMeasure{
 				if( addInfo ){
 					layout.setDivisionsStyle(painter,false);
 
-					String repetitions = ("x" + this.getRepeatClose());
+					String repetitions = ("x" + (this.getRepeatClose() + 1));
 					painter.drawString(repetitions, x2 - painter.getFMWidth(repetitions) + getSpacing() - size, y1 + painter.getFMBaseLine() - (2f * scale));
 				}
 			}
--- a/desktop/TuxGuitar-nsm/jni/GNUmakefile
+++ b/desktop/TuxGuitar-nsm/jni/GNUmakefile
@@ -1,0 +1,14 @@
+# See pom.xml for variable definitions
+
+OBJECTS=app_tuxguitar_nsm_NSMSignal.o
+
+%.o: %.c
+	$(CC) $(CFLAGS) -c -o $@ $<
+
+all:    $(LIBRARY)
+
+$(LIBRARY):	$(OBJECTS)
+	$(CC) $(LDFLAGS) -shared -o $(LIBRARY) $(OBJECTS) $(LDPATH) $(LDLIBS)
+
+clean:
+	$(RM) $(OBJECTS) $(LIBRARY)
--- a/desktop/TuxGuitar-nsm/jni/app_tuxguitar_nsm_NSMSignal.c
+++ b/desktop/TuxGuitar-nsm/jni/app_tuxguitar_nsm_NSMSignal.c
@@ -16,6 +16,31 @@ static void sigterm_handler(int sig) {
     }
 }
 
+static int install_sigterm_handler(void) {
+    struct sigaction sa;
+    memset(&sa, 0, sizeof(sa));
+    sa.sa_handler = sigterm_handler;
+    sigemptyset(&sa.sa_mask);
+    sa.sa_flags = SA_RESTART;
+    return sigaction(SIGTERM, &sa, NULL);
+}
+
+/*
+ * Class:     app_tuxguitar_nsm_NSMSignal
+ * Method:    reinstallSigtermHandler
+ * Signature: ()V
+ *
+ * Re-installs our sigterm_handler without creating a new pipe.  Call this
+ * just before sending any NSM reply so our handler is the last one installed
+ * (libjack overrides it when FluidSynth connects to JACK).
+ */
+JNIEXPORT void JNICALL
+Java_app_tuxguitar_nsm_NSMSignal_reinstallSigtermHandler(JNIEnv *env, jclass cls) {
+    if (sigterm_pipe_write != -1) {
+        install_sigterm_handler();
+    }
+}
+
 /*
  * Class:     app_tuxguitar_nsm_NSMSignal
  * Method:    registerSigtermPipe
@@ -33,12 +58,7 @@ Java_app_tuxguitar_nsm_NSMSignal_registerSigtermPipe(JNIEnv *env, jclass cls) {
     }
     sigterm_pipe_write = fds[1];
 
-    struct sigaction sa;
-    memset(&sa, 0, sizeof(sa));
-    sa.sa_handler = sigterm_handler;
-    sigemptyset(&sa.sa_mask);
-    sa.sa_flags = SA_RESTART;
-    if (sigaction(SIGTERM, &sa, NULL) != 0) {
+    if (install_sigterm_handler() != 0) {
         close(fds[0]);
         close(fds[1]);
         sigterm_pipe_write = -1;
--- a/desktop/TuxGuitar-nsm/jni/app_tuxguitar_nsm_NSMSignal.c
+++ b/desktop/TuxGuitar-nsm/jni/app_tuxguitar_nsm_NSMSignal.c
@@ -1,0 +1,63 @@
+#include <jni.h>
+#include <signal.h>
+#include <unistd.h>
+#include <string.h>
+
+/* Write end of the self-pipe; -1 means not initialised. */
+static volatile int sigterm_pipe_write = -1;
+
+static void sigterm_handler(int sig) {
+    (void)sig;
+    int fd = sigterm_pipe_write;
+    if (fd != -1) {
+        /* write() is async-signal-safe; ignore partial-write — one byte is enough. */
+        char b = 1;
+        write(fd, &b, 1);
+    }
+}
+
+/*
+ * Class:     app_tuxguitar_nsm_NSMSignal
+ * Method:    registerSigtermPipe
+ * Signature: ()I
+ *
+ * Creates a pipe, installs our sigterm_handler for SIGTERM (overriding any
+ * handler previously set by SWT/GTK/FluidSynth), and returns the read-end fd.
+ * Returns -1 on error.
+ */
+JNIEXPORT jint JNICALL
+Java_app_tuxguitar_nsm_NSMSignal_registerSigtermPipe(JNIEnv *env, jclass cls) {
+    int fds[2];
+    if (pipe(fds) != 0) {
+        return -1;
+    }
+    sigterm_pipe_write = fds[1];
+
+    struct sigaction sa;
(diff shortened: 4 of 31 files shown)


ycollet:
Here is a sum-up of the NSM protocol:
```
NSM Protocol — Requirements Summary

Transport

- OSC messages over UDP
- Server URL provided via the $NSM_URL environment variable (osc.udp://host:port)

Startup — /nsm/server/announce

Client sends on startup:
/nsm/server/announce  s:appName  s:capabilities  s:executableName  i:apiMajor  i:apiMinor  i:pid
- pid must be the PID the session manager should send SIGTERM to (our bash wrapper PID via $NSM_LAUNCHER_PID)
- capabilities is a colon-delimited string (e.g. :dirty:switch:); we currently declare none ("")

Server replies with /reply /nsm/server/announce on success, or /error on failure.

Session open — /nsm/client/open

Server sends a path (the client's private session directory):
/nsm/client/open  s:sessionDir  s:displayName  s:clientId
Client must:
1. Create the directory if absent
2. Load its state from that directory
3. Reply: /reply /nsm/client/open  s:"opened"  (or /error on failure)

Session save — /nsm/client/save

/nsm/client/save
Client must write all state to the session directory, then reply:
/reply /nsm/client/save  s:"saved"

Quit — /nsm/client/quit  (or SIGTERM)

NSM may either send /nsm/client/quit over OSC or send SIGTERM to the announced PID. Client must close cleanly in both cases.

Session loaded — /nsm/client/session_is_loaded

After all clients have replied to /nsm/client/open, the server broadcasts this to all clients. It signals that the session is fully loaded (all peers are ready).

Error replies

/error  s:path  i:errorCode  s:message

Key invariants

┌───────────────────────────────────────────────┬───────────────────────────────┐
│                     Rule                      │     Consequence if broken     │
├───────────────────────────────────────────────┼───────────────────────────────┤
│ Reply to every /open and /save                │ Session manager hangs waiting │
├───────────────────────────────────────────────┼───────────────────────────────┤
│ Announce before any other message             │ Server ignores late clients   │
├───────────────────────────────────────────────┼───────────────────────────────┤
│ Write ALL state to the session dir            │ State lost on session switch  │
├───────────────────────────────────────────────┼───────────────────────────────┤
│ Tolerate SIGTERM at any time after save reply │ Session close fails           │
└───────────────────────────────────────────────┴───────────────────────────────┘
```

Houston4444:
Hi !
Quick tests this morning. I can see 3 problems (by importance order).

* You have broken the regular SIGINT signal behavior. `tuxguitar` quits normally under NSM, but if program is started normally in a terminal, Ctrl+C stops the process in the terminal but TuxGuitar window does not close and is still usable. 
* The splash screen should be hidden, and tuxguitar replies to the /nsm/client/open message too quickly, it should reply when it is ready to work, not during the splash screen.
* The .desktop file should contains `X-NSM-Capable=true`, this way TuxGuitar can be listed in applications

That's a lot of code, I hope you understand exactly what each line does.

Houston4444:
Thanks, it is better now. I forgot an important thing.
The actions "new", "open" and "save as" should be disabled or make an action coherent with the session managing. For example, "Save as" is acceptable if it is just a backup and we stay working on the project (without switch to new saved file). Same for "New" and "Open", acceptable only if it replaces the current project with the choose one (as QTractor does), else it must be disabled.

It make me think that  TuxGuitar should not be able to load many tabs under NSM. Anyway, when we re-open it, there is only one tab. Then the ':switch:' NSM capability could help to switch faster from one project to another.

ycollet:
Thanks a lot for the tests. I will check if I can improve the PR.

Houston4444:
About tabs, I don't know what could be the good choice. TuxGuitar is one instance only, so forbid tabs also forbids some files interactions such as copy/paste a measure.

Note that the way you've done it, the saved .tg file is the current .tg file, it can be not exactly what user could want. Bref, this point needs attention and needs a (not easy) choice.

ycollet:
For the NSM session, my usage of Tuxguitar during a NSM session:
I will open a tablature which is often outside of the NSM path. During a session, I can change the score (fingering, track level) and may want to save these changes outside the NSM path. I also may want to create a new score and copy sections of a score already opened in tuxguitar during this NSM session. So, to me, we should not disable New and Open and the Save As can be allowed to save a file outside the NSM path.

So, for me:
- New / Open / Save As: leave them completely untouched — no interception, no restrictions.
- NSM save: instead of always writing to {sessionDir}/tuxguitar.tg, save the current document to its actual path (wherever TuxGuitar thinks the file is). If it's a brand-new unsaved document, fall back to the session directory. After saving, write a small pointer file in the session directory recording which file was saved.
- NSM open (session restore): read the pointer file from the session directory to find which file to reopen. If the pointer file doesn't exist (first launch in this session), start with an empty template as today.

The main tradeoff vs. the current approach: the session directory no longer holds the authoritative copy of the song — it just holds a reference to wherever the user actually saved. This means if the user moves or deletes the external file, session restore will fail gracefully (fall back to empty). But for your workflow this is the right behaviour.


----- Mail original -----
De: "Houston4444" ***@***.***>
À: "helge17/tuxguitar" ***@***.***>
Cc: "ycollet" ***@***.***>, "Author" ***@***.***>
Envoyé: Mardi 9 Juin 2026 14:57:18
Objet: Re: [helge17/tuxguitar] Add NSM support (PR #1080)


Houston4444 left a comment (helge17/tuxguitar#1080) 

About tabs, I don't know what could be the good choice. TuxGuitar is one instance only, so forbid tabs also forbids some files interactions such as copy/paste a measure. 

Note that the way you've done it, the saved .tg file is the current .tg file, it can be not exactly what user could want. Bref, this point needs attention and needs a (not easy) choice. 

— 
Reply to this email directly, view it on GitHub , or unsubscribe . 
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS and Android . Download it today!

Houston4444:
There is another major problem: it is not possible to start tuxguitar under NSM if tuxguitar is already started.

When I stop the tuxguitar client in RaySession with unsaved modifications (in the current song or another one), tuxguitar asks with a dialog if we want to save changes. This is strictly forbidden, the program MUST stop ASAP without saving.

Theses 2 major problems make me say that a new instance force when run under NSM would be a good choice. the only alternative choice I see would be to have a proxy process (with OSC communication with the NSM server, and that can communicate with the main tuxguitar process).

Note that I have not tested how the tuxguitar JACK client works (I don't know how to make it works).    

Your approach with file actions seems correct, the essential is that:
* all config files needed to restore the client are in the session folder (symlinks are ok of course)
* The song we save is the song that will be open at next session start

Houston4444:
I suggest an approach (if it is technically possible). Under NSM, tuxguitar should start a new instance, and at NSM save, it should save all the open songs, it should also save in a file which files are open, then when we open tuxguitar under NSM, all the songs are open.

guiv42:
Let me be honest, I have no idea about how to handle this PR. I clearly understand the feature has an added value, but there are too many reasons why I'm really uncomfortable:
- I am currently completely incapable of reviewing it: I don't know how to test it, I'm not even clear about what *precisely* it is *supposed to do*
- PR is significant in size (1k+ lines of code, without any test). Review requires an effort I cannot currently afford (limited available time)
- I don't understand why the OSC and NSM layers are not clearly separated
- it seems the code creates new threads, without any consideration for the existing threads model (TuxGuitar manages a pool of threads created once at init). No idea of the consequences.
- it's not trivial (at least *for me*): it manipulates patterns that are not so widely used in the code base. E.g. JNI, `EarlyInitPlugin` (concept was created for a macOS-specific case). And I don't understand why. It also involves sw layers I have never looked at (interface with OS for signal handling)
- if any issue pops up one day, I would be completely incapable of analyzing, troubleshooting, of fixing anything
- I must admit that the use of AI-assisted coding certainly doesn't bring any confidence to me (it's rather the opposite)

In other words, I am very, *very* far from "understanding exactly what each line does", as mentioned in a message above.

An important reminder about the context of TuxGuitar development: we inherited the code base from its original repo without any support from its author, with zero documentation and zero tests. Architecture was never described anywhere (as far as I know). We've been learning since the creation of this fork about 3 years ago, but this context explains why we are globally progressing at a very slow pace.

So I hope you will understand I just cannot merge something with such a low level of confidence.

I suggest leaving this PR open, but I'm definitely not optimistic about my capacity to merge (or even review) this in any near future.

Houston4444:
@guiv42 , thanks for these words. I completely understand your position, I think I would feel the same way if I were in your shoes. So, I'll try to answer point by point

> I am currently completely incapable of reviewing it: I don't know how to test it, I'm not even clear about what precisely it is supposed to do

I could explain that with precision on demand, to test it we need to run tuxguitar in RaySession and/or New-Session-Manager and check the behavior is correct and follows the API rules.

> I don't understand why the OSC and NSM layers are not clearly separated

I can't talk myself about how it works in this PR, but the OSC / NSM separation is not necessarily a good idea. NSM uses OSC to communicate with the NSM server, if in future TuxGuitar uses OSC for other remotes (it can be a great feature of course), use the same OSC server for NSM and other controls is a very good point (not at all required, but it allows easier OSC communication with brother clients in the NSM session).

> I must admit that the use of AI-assisted coding certainly doesn't bring any confidence to me (it's rather the opposite)

I can't disagree with you on this point. The few times I have tested it, I must admit the result is near to be formidable very quickly, but if we look in the code more clearly there are some unintelligible errors, massive use of deprecated methods, etc... Finally, it takes time to fix errors and it leaves one feeling constantly uneasy. Finally I use it only to ask it some questions (sometimes with big success I would have never found with a search engine), and that's all.

I can't answer myself to other questions, I would be happy to help, understand and discuss what this PR technically does exactly, but I would need to learn JAVA and learn how is built this repository (it seems still hard for you ;)).

So, I'm afraid we're stuck in a sad status quo.

guiv42:
About OSC and NSM: having an independent OSC implementation would ease the validation (it could be tested independently). That could also make the feature available in other contexts (or other projects). Re-using an existing and "proven in use" library would have been much more comfortable. No idea if this kind of re-usable object exists for OSC.

> I suggest an approach (if it is technically possible). Under NSM, tuxguitar should start a new instance

I have a good understanding of this part of the code, and that would be simple to implement. But this example illustrates very well the complexity of the situation: I don't know at all **why** the implementation explicitly ensures there's only 1 running instance. There might be side effects that I'm really not aware of. An example of scenario could be an unprotected parallel and conflicting access to some resource by multiple instances (e.g. config file, connection to the MIDI system, etc).

ycollet:
I updated the PR with:
- move OSC into a separate module
- add tests to this OSC module.
I think tests for the NSM part are missing, I will add these tomorrow.

For my TuxGuitar / NSM usage: I load guitarpro files from outside the NSM directory and from time to time, I adjust things in the score and I want to be able to export these scores as GP5 files. So, I think it's better not to disable the open / save / save as buttons.
At least, if the original PATH to the file opened in tuxguitar doesn't exist anymore, then we can save a backup tg file in the NSM directory.

guiv42:
Like other AI-generated PRs, I close this one as is: sorry, but I do not see any credible path to get an acceptable level of confidence.

This implementation can still be useful, if someone motivated and skilled can reverse-engineer it, understand what it's trying to do, and re-implement it in accordance with TuxGuitar current design. That's out of reach for me.

PR #1082, opened 2026-05-27
feat: persist and restore zoom level across sessions

drewmerc302:
## Summary
- Adds `layout.scale` config key (default `1.0`) to save the tablature zoom level
- Zoom is persisted to config on every zoom change and on shutdown
- On startup, saved zoom is restored and clamped to the valid 0.5–2.0 range

## Files changed
- `TGConfigKeys.java` — new `LAYOUT_SCALE` constant
- `TGConfigDefaults.java` — default value `"1.0"`
- `TGSetLayoutScaleAction.java` — persist zoom on change
- `TGDisposeAction.java` — persist zoom on shutdown
- `TablatureEditor.java` — restore zoom on startup

## Test plan
- [ ] Set zoom to a non-default level (e.g. 1.5x), close TuxGuitar, reopen — zoom should restore
- [ ] Set zoom to extreme values near 0.5 or 2.0, verify clamp works on restore
- [ ] Fresh install with no config — should default to 1.0x

### DIFF
--- a/desktop/TuxGuitar/src/app/tuxguitar/app/action/impl/layout/TGSetLayoutScaleAction.java
+++ b/desktop/TuxGuitar/src/app/tuxguitar/app/action/impl/layout/TGSetLayoutScaleAction.java
@@ -1,6 +1,8 @@
 package app.tuxguitar.app.action.impl.layout;
 
 import app.tuxguitar.action.TGActionContext;
+import app.tuxguitar.app.system.config.TGConfigKeys;
+import app.tuxguitar.app.system.config.TGConfigManager;
 import app.tuxguitar.app.view.component.tab.Tablature;
 import app.tuxguitar.app.view.component.tab.TablatureEditor;
 import app.tuxguitar.editor.action.TGActionBase;
@@ -21,5 +23,9 @@ public class TGSetLayoutScaleAction extends TGActionBase{
 
 		Tablature tablature = TablatureEditor.getInstance(getContext()).getTablature();
 		tablature.scale(scale);
+
+		// Persist zoom level to config
+		TGConfigManager config = TGConfigManager.getInstance(getContext());
+		config.setValue(TGConfigKeys.LAYOUT_SCALE, tablature.getScale());
 	}
 }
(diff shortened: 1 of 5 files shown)


desktop/TuxGuitar/src/app/tuxguitar/app/view/component/tab/TablatureEditor.java
```diff
@@ -27,6 +29,13 @@ public TablatureEditor(TGContext context) {
 
 	public void initialize() {
 		this.tablature = new Tablature(this.context, TGDocumentManager.getInstance(this.context));
+
+		// Restore saved zoom level from config, clamped to valid range
+		TGConfigManager config = TGConfigManager.getInstance(this.context);
+		float savedScale = config.getFloatValue(TGConfigKeys.LAYOUT_SCALE, Tablature.DEFAULT_SCALE);
+		savedScale = Math.max(0.5f, Math.min(savedScale, 2.0f));
```
guiv42, line 36: suggestion: replace absolute values by `static final` constants, like `MIN_SCALE` and `MAX_SCALE`

desktop/TuxGuitar/src/app/tuxguitar/app/action/impl/layout/TGSetLayoutScaleAction.java
```diff
@@ -21,5 +23,9 @@ protected void processAction(TGActionContext context) {
 
 		Tablature tablature = TablatureEditor.getInstance(getContext()).getTablature();
 		tablature.scale(scale);
+
+		// Persist zoom level to config
```
guiv42, line 27: I see no reason to persist zoom level here, I'd say it's sufficient to do it when closing app (what you did correctly in `TGDisposeAction`).
The only case I see where this could be useful is if the app crashes at some point, then it the latest config value may not be saved. Did I miss something?

I don't think it's really useful, and I would prefer to limit the "noise" on the filesystem (e.g. when zooming in/out with 'Ctrl+mouse wheel' it generates many write requests).
So I suggest to revert the modifications in that file.

guiv42 (commented):
just a few minor remarks, nothing critical

olaquetal:
@drewmerc302 seems interesting!
Does it work also on android? (I'm mostly an android user)
Thanks

helge17:
> Does it work also on android?

Currently, this PR only applies to the desktop.

guiv42:
@drewmerc302 : I have merged your commit and applied the few suggested corrections through another PR (#1098), so I close this one.
Thanks for your contribution!

PR #1083, opened 2026-05-27
fix: resolve dark mode color issues in track table

drewmerc302:
## Summary
- `SWTAppearance.java`: `WidgetHighlightBackground` was mapped to `COLOR_WIDGET_HIGHLIGHT_SHADOW` alone, which resolves to white in dark mode — making even rows in the track table white-on-white. Changed to blend `WIDGET_BACKGROUND + WIDGET_NORMAL_SHADOW` for a visible midtone (same pattern already used by `WidgetSelectedBackground` on the adjacent line).
- `TGTableColorModel.java`: Selected row used `WidgetSelectedBackground` (a muted gray blend), swapped to `InputSelectedBackground` which picks up the system accent color for a clearer selection highlight in both light and dark themes.

## Files changed
- `SWTAppearance.java` — fix even-row highlight color
- `TGTableColorModel.java` — use system accent for selected row

## Test plan
- [ ] Open a multi-track song in dark mode — even/odd row striping should be visible
- [ ] Selected track row should use system accent color (e.g. blue on macOS)
- [ ] Verify no regression in light mode — rows and selection still distinguishable

### DIFF
--- a/desktop/TuxGuitar-ui-toolkit-swt/src/app/tuxguitar/ui/swt/appearance/SWTAppearance.java
+++ b/desktop/TuxGuitar-ui-toolkit-swt/src/app/tuxguitar/ui/swt/appearance/SWTAppearance.java
@@ -26,7 +26,7 @@ public class SWTAppearance implements UIAppearance {
 		this.colorMap.put(UIColorAppearance.WidgetForeground, this.createColorModel(SWT.COLOR_WIDGET_FOREGROUND));
 		this.colorMap.put(UIColorAppearance.WidgetLightBackground, this.createColorModel(SWT.COLOR_WIDGET_BACKGROUND, SWT.COLOR_WIDGET_HIGHLIGHT_SHADOW));
 		this.colorMap.put(UIColorAppearance.WidgetLightForeground, this.createColorModel(SWT.COLOR_WIDGET_FOREGROUND));
-		this.colorMap.put(UIColorAppearance.WidgetHighlightBackground, this.createColorModel(SWT.COLOR_WIDGET_HIGHLIGHT_SHADOW));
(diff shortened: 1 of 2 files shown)


guiv42:
Colors are always somehow subjective. To be fully honest:
- `SWTAppearance`: maybe it's linked to my config, but I don't see a big difference in dark mode. I'd say it is slightly more readable in light mode with this PR
- I'm not enthusiastic about the use of `InputSelectedBackground/Foreground` for `SELECTED_LINE`. Using a color dedicated to an "input" field does not really make sense here, and it may give weird results in some environments. The result in my configuration is a blue that I find a bit aggressive.

@helge17 , any opinion about this?

Just an example in my configuration (Linux opensuse tumbleweed, KDE, X11), with current master branch:
light_ref
dark_ref

and with this PR:
light_pr
dark_pr

@drewmerc302 : that should probably be useful to add here some screenshots in your environment.

helge17:
On Debian 13 KDE, TuxGuitar looks the same as shown above.

Here are a few more snapshots.

I think the patch improves readability, I would suggest to merge it. The blue is indeed very light, but it’s also used for other GUI elements, also in LibreOffice and KDE. In Kmail, the mail list looks quite similar to the track table in TuxGuitar with this patch (see below).

FreeBSDGnomeMasterBright:
FreeBSDGnomeMasterBright

FreeBSDGnome1083Bright:
FreeBSDGnome1083Bright

MacMasterBright:
MacMasterBright

Mac1083Bright:
Mac1083Bright

MacMasterDark:
MacMasterDark

Mac1083Dark:
Mac1083Dark

WinMasterBright:
WinMasterBright

Win1083Bright:
Win1083Bright

Deb13KDEKmailBright:
Deb13KDEKmailBright

Deb13KDEKmailDark:
Deb13KDEKmailDark

guiv42:
OK, fine, thanks for all these tests. Especially the "MacMasterDark" configuration clearly shows an issue with current master branch.
So let's go for a merge.

PR #1096, opened 2026-06-14
always stop sequencer when stopping player

### LINKED ISSUE #1092: bug: playing 2 songs concurrently
**Configuration**: TuxGuitar 2.0.1, Linux (openSuse tumbleweed), TuxGuitar installed from openSuse repository
**Sound configuration** : menu "tools / settings", then "sound", then select "TuxGuitar Sequencer". Bug is not reproduced with "Real Time Sequencer"

**Detailed procedure to reproduce**:
- open 2 different songs in TuxGuitar
- activate "Count in" (via "Player" menu or main toolbar)
- click on "play" button
- while count-in is still in progress, click on "stop" button
- switch to the second song
- click on "play" button

**Expected behavior**: second song is played
**Observed behavior**: the 2 songs are played simultaneously

guiv42:
should fix #1092

While playing count-in, sequencer is initialized, but not yet in "running" state.
If player was stopped after count-in started, but before first measure starts, then the sequencer was never reset.

Bug scenario fixed by this commit:
- open several songs, activate count-in
- start playing
- stop while count-in is played (before playing first measure)
- switch song
- start playing

With TuxGuitar Sequencer, the midi sequence from first song is never reset. So, the 2 songs are played simultaneously

Draft pull request: not tested in all configurations.
Tested: TuxGuitar Sequencer, Real Time Sequencer
Not tested: Jack sequencer

### DIFF
--- a/common/TuxGuitar-lib/src/main/java/app/tuxguitar/player/base/MidiPlayer.java
+++ b/common/TuxGuitar-lib/src/main/java/app/tuxguitar/player/base/MidiPlayer.java
@@ -115,10 +115,7 @@ public class MidiPlayer{
 	public void stopSequencer() {
 		try {
 			this.lock();
-
-			if( this.getSequencer().isRunning() ){
-				this.getSequencer().stop();
-			}
+			this.getSequencer().stop();
 		} catch (MidiPlayerException e) {
 			e.printStackTrace();
 		} finally {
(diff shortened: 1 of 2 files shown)


guiv42:
tested with Jack sequencer: bug was present also, and is fixed by this PR.

To be merged after 2.1.0 is released

guiv42:
Don't want to leave too many PR open until 2.1.0 is released => fix merged in tuxguitar-next branch by #1122

Issue #554, opened 2024-10-18
During playback, bars get rearranged on the screen


buryoutougen:
It's hard to come up with a specific example to reproduce, because it depends on the tab contents and probably also on your OS, screen resolution etc., but basically, try to open any larger tab and start playback from different bars. Chances are the layout of the bars will change, with the bar you're in jumping to a different part of the screen.

I think this is caused by the removal of the scrollbar during playback in one of the updates.

guiv42:
@buryoutougen : could you please precise your configuration? Especially the version of TuxGuitar you are using.
I fixed recently a scrolling issue that was introduced together with the scrollbars removal (#527), it could possibly be linked to your problem.

buryoutougen:
I tried 2024-08-02-master and 2024-09-28-master (SWT version on Windows 11, 1920x1080 resolution). The behavior is the same.

buryoutougen:
I recorded my screen to illustrate it better. I'm pressing spacebar (play/pause).

https://github.com/user-attachments/assets/809cf313-d425-4419-91e9-53e39cae8cda

guiv42:
Thanks for the precision. The fix I mentioned is already present in 2024-09-28 snapshot, so your problem is something different.
I hope I won't have to revert the toolbar removal, these toolbars gave me a lot of headaches already (#403)

guiv42:
I've possibly found a solution for the *clean* option (#557).
It's a rather aggressive performance optimization, with the objective to fix #403 while keeping scrollbar(s) displayed when playing. Strategy is to repaint only what is needed / when it's needed instead of repainting full tab systematically. App profiling in my test case shows significant improvement, this is encouraging.
It deserves to be tested more deeply in next pre-release of master branch, there is a possible risk of regression I think (e.g. tab not refreshed in corner cases?).
Note: optimization is only applied when playing.

guiv42:
Fix is available from pre-release [2024-10-23](https://github.com/helge17/tuxguitar/releases/tag/2024-10-23-master), **in master branch**

buryoutougen:
Works perfectly for me, thanks!

PR #1099, opened 2026-06-17
Fix JFX number-key note entry on startup (fixes #1005)

### LINKED ISSUE #1005: JFX: Entering notes using the number keys not working
After starting TuxGuitar (current master branch), it is not possible to enter notes using the number keys on the keyboard. Only after klicking into the track table below the tablature or after hitting the arrow down key, the number keys start to work normally.

Same behaviour on all desktop OS with JFX. SWT works fine.

The error seems to be related to PR #544, commit 96012364c84421a3b67aa42363f40b2e20384856.

advit-singh:
JFXImageView.redraw() incorrectly called requestFocus(), which pulled keyboard focus away from the tab editor on toolbar layout.

Summary
- Remove error-causing requestFocus() from JFXImageView.redraw().
- Toolbar ImageView layout no longer steals keyboard focus from the tab editor.

Fixes #1005

### DIFF
--- a/desktop/TuxGuitar-ui-toolkit-jfx/src/app/tuxguitar/ui/jfx/widget/JFXImageView.java
+++ b/desktop/TuxGuitar-ui-toolkit-jfx/src/app/tuxguitar/ui/jfx/widget/JFXImageView.java
@@ -56,7 +56,6 @@ public class JFXImageView extends JFXNode<ImageView> implements UIImageView {
 	}
 
 	public void redraw() {
-		this.getControl().requestFocus();
 	}
 
 	public void addResizeListener(UIResizeListener listener) {


guiv42:
Can you please provide some more detailed explanation?
I don't see why the call to `requestFocus` would be erroneous (it was probably put there for some reason). Yet I have no idea of the consequences of this commit.

In fact, I would like to understand first what is the precise cause of bug #1005. It seems it was introduced when I re-implemented the main toolbar, and to be honest I really don't understand what's going on. Note that I did not have the required time to analyze this specific bug (JFX is not a priority).
In other words: I need to understand what causes the bug, why the proposed modification fixes it, and all the possible consequences of that modification. Just testing that it fixes the bug, without a detailed understanding, is not sufficient for me to merge this PR.

advit-singh:
**Why the Bug Happened:**

On startup, `TGTabFolder.updateFocus()` tries to put focus on the tab editor canvas. After PR #554, the toolbar includes a tempo indicator (built with `UIImageView`). When window is laid out, the ImageView `setbounds()` calls `JFXImageView.redraw()`

Before this fix, the focus went to `ImageView`, as the `ImageView` introduced by PR #554 steals the focus from the editor. Since this tempo `ImageView` doesn't have any key-bind listeners, the number keys don't do anything when this is in focus.

**Current Fix:**

The `UIControl` interface defines `redraw()` for repainting, and the implementation is different per widget:
- `JFXCanvas` fires the paint listeners
- `JFXRegion` calls `requestLayout()`
- SWTControl calls a native `widget.redraw()`

Before this fix, the `redraw()` in `JFXImageView` called `requestFocus()` only, which isn't a redraw, just taking focus.

For the JFX `ImageView`, the image and layout updates are handled via `setImage()` and `setBounds()` respectively, so the `redraw()` isn't needed. The `requestFocus()` call has been in `JFXImageView` since the JFX port; the toolbar rework didn't add it in. The toolbar re-implementation just surfaced the issue by laying out a `ImageView` in the toolbar.

**Consequences (aside from expected bug fix):**
- no expected visual regression, since the icon still updates through `setImage()`, and other `UIImageView` uses also utilize `setImage()` for content
- scope: only in JFX, not Android or SWT

---

Hope this helps, please let me know if anything else is needed.

guiv42:
Yes it helps, thank you very much for this precise explanation. Your analysis is confirmed by the following tests (done on 28463804718585ffd70e2e9145ad860856fcb16d):
- launch TuxGuitar from command line, with an argument pointing to a file with tempo 96 bpm: known issue, the tablature does not have the focus, notes cannot be created with keyboard
- put manually the focus on the tablature, check that notes can be created with keyboard
- type F9 and select tempo 110%, then play: tablature has lost the focus. Because main toolbar was redrawn (a89111bb0f682218dc21786af95133b8d2dd0c31)
- restore manually the focus on the tablature
- click on the button at the right of the main toolbar (toolbar configuration), then click OK in the dialog: main toolbar is redrawn and tablature looses focus again
- restore manually the focus on the tablature
- edit the toolbar configuration again, and remove tempo indicator from toolbar. After clicking OK, the toolbar is redrawn, but the tablature still has the focus and notes can be created normally

What I still don't understand is this call to `requestFocus()` in a `redraw()` method. As it does not seem logical, I suspect there may be some hidden reason for that, but who knows. It's been there for 10 years, no idea why.

The good news is that the proposed patch is strictly restricted to JFX configuration, with zero impact on the official releases (where only SWT variant is published).

So, my proposal is to wait until 2.1.0 is stabilized and released, and then to merge this PR as is.

Thanks for your contribution.

guiv42 (approved):
OK, to be merged after 2.1.0 release

PR #1110, opened 2026-07-09
add config to allow elapsed time to continue accumulating during loop

uaihebert:
Hello everyone, this is my first PR here.

I tried to keep the changes as small as possible.

I added a configuration to allow the elapsed time to increase.

Imagine that a person wants to do the same exercise for 5minutes, but if the elapsed time resets, it makes it harder to keep track of.
image
This is how it looks when a player opens if the first time.
image
It is still possible to play looping with elapsed time resetting (default behaviour today).
image
image
Let me know if this is something desirable and if I need to make any other changes to keep the solution with the desired code structure.

### DIFF
--- a/common/TuxGuitar-lib/src/main/java/app/tuxguitar/player/base/ElapsedTimeLoopAccumulator.java
+++ b/common/TuxGuitar-lib/src/main/java/app/tuxguitar/player/base/ElapsedTimeLoopAccumulator.java
@@ -1,0 +1,25 @@
+package app.tuxguitar.player.base;
+
+public class ElapsedTimeLoopAccumulator {
+
+	private long baseElapsedTime;
+	private long lastTimestamp;
+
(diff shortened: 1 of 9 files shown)


guiv42:
Hi
still sharing my opinion about this, even if you have closed this PR.

Take care, elapsed time indicator is pretty tricky. It's not obvious exactly *what* this indicator represents. Especially, it does not necessarily progress with real time: it is impacted by tempo percentage. E.g. when playing at tempo = 50%, the elapsed time indicator increases by 1 second every 2 seconds.

In the most classic use case (just playing a song), this indicator should be consistent with a "song timestamp": it tells you where you are in the song.
It becomes tricky in 2 cases: when there are repeats, and when there is a loop. Combining the 2 makes it seriously complex. Especially when the loop does not cover the full repeated part!

PR #1111, opened 2026-07-09
add config to allow elapsed time to continue accumulating during loop

uaihebert:
Hello everyone, this is my first PR here.

I tried to keep the changes as small as possible.

I added a configuration to allow the elapsed time to increase.

Imagine that a person wants to do the same exercise for 5minutes, but if the elapsed time resets, it makes it harder to keep track of.
image
This is how it looks when a player opens it for the first time.
image
It is still possible to play looping with elapsed time resetting (default behaviour today).
image
image
Let me know if this is something desirable and if I need to make any other changes to keep the solution with the desired code structure.

### DIFF
--- a/common/TuxGuitar-lib/src/main/java/app/tuxguitar/player/base/ElapsedTimeLoopAccumulator.java
+++ b/common/TuxGuitar-lib/src/main/java/app/tuxguitar/player/base/ElapsedTimeLoopAccumulator.java
@@ -1,0 +1,25 @@
+package app.tuxguitar.player.base;
+
+public class ElapsedTimeLoopAccumulator {
+
+	private long baseElapsedTime;
+	private long lastTimestamp;
+
+	public ElapsedTimeLoopAccumulator() {
+		this.reset();
+	}
+
+	public long accumulate(long currentTimestamp) {
+		if (currentTimestamp < this.lastTimestamp) {
+			this.baseElapsedTime += this.lastTimestamp;
+		}
+
+		this.lastTimestamp = currentTimestamp;
+		return (this.baseElapsedTime + currentTimestamp);
+	}
+
+	public void reset() {
+		this.baseElapsedTime = 0;
+		this.lastTimestamp = 0;
+	}
+}
--- a/common/TuxGuitar-lib/src/main/java/app/tuxguitar/player/base/MidiPlayer.java
+++ b/common/TuxGuitar-lib/src/main/java/app/tuxguitar/player/base/MidiPlayer.java
@@ -427,7 +427,7 @@ public class MidiPlayer{
 		return false;
 	}
 
-	private boolean isPaused() {
+	public boolean isPaused() {
 		try {
 			this.lock();
 
--- a/common/TuxGuitar-lib/src/test/java/app/tuxguitar/player/base/TestElapsedTimeLoopAccumulator.java
+++ b/common/TuxGuitar-lib/src/test/java/app/tuxguitar/player/base/TestElapsedTimeLoopAccumulator.java
@@ -1,0 +1,28 @@
+package app.tuxguitar.player.base;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+public class TestElapsedTimeLoopAccumulator {
+
+	@Test
+	public void onAccumulateWhenTimestampWrapsExpectContinuousElapsedTime() {
+		ElapsedTimeLoopAccumulator accumulator = new ElapsedTimeLoopAccumulator();
+
+		assertEquals(120L, accumulator.accumulate(120L));
+		assertEquals(380L, accumulator.accumulate(380L));
+		assertEquals(430L, accumulator.accumulate(50L));
+	}
+
+	@Test
+	public void onResetWhenNewPlaybackStartsExpectElapsedTimeStartsFromLoopTimestamp() {
+		ElapsedTimeLoopAccumulator accumulator = new ElapsedTimeLoopAccumulator();
+
+		accumulator.accumulate(900L);
+		accumulator.accumulate(20L);
+		accumulator.reset();
+
+		assertEquals(40L, accumulator.accumulate(40L));
+	}
+}
--- a/common/resources/lang/messages.properties
+++ b/common/resources/lang/messages.properties
@@ -588,6 +588,7 @@ transport.mode=Play Mode
 transport.mode.simple=Simple Mode
 transport.mode.simple.tempo-percent=Percentage of Tempo
 transport.mode.simple.loop=Play looped
+transport.mode.elapsed-time.dont-stop-during-loop=Don't reset "Elapsed Time" during loops
 transport.mode.trainer=Training Mode
 transport.mode.trainer.increment-description=Increment by
 transport.mode.loop-range=Loop Range
--- a/common/resources/lang/messages_pt.properties
+++ b/common/resources/lang/messages_pt.properties
@@ -445,6 +445,7 @@ transport.mode=Modo Tocador
 transport.mode.simple=Modo Tocador Simples
 transport.mode.simple.tempo-percent=Percentual de tempo
 transport.mode.simple.loop=Tocar um Loop
+transport.mode.elapsed-time.dont-stop-during-loop=Não reiniciar "Tempo Decorrido" durante loops
 transport.mode.trainer=Modo de Treinamento
 transport.mode.trainer.increment-description=Incrementar em
 transport.mode.loop-range=Tamanho do loop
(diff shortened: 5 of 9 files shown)


uaihebert:
hello @guiv42 , i saw your comment on the old PR: https://github.com/helge17/tuxguitar/pull/1110#issuecomment-4922971695

uaihebert:
i did not know about the tempo changing the elapsed time. indeed, it may show wrong data then. ty for the feedback.

uaihebert:
i will change the approach then, and open a new PR. 

is this something desired? if yes, I thought about putting it in position 1. @guiv42 

image

but I could add it as a command and to be used in position 2 .

guiv42:
> i will change the approach then, and open a new PR.

Can you please provide more details before implementing?
As I mentioned in #1110 it's not so obvious to me what this time counter is *supposed to display*.
So, to be honest I'm not enthusiastic about changing the actual behavior. It would probably be useful to discuss what you want to do before implementing it (just to save time!)

And no, I don't think it's a good idea to move the time counter to the left toolbar. Reason for this is simple: the main toolbar is completely configurable, user can choose to display time counter or not (and where). Left toolbar is currently hard-coded.
That would be more useful to make the left toolbar configurable also, but that's a significant job.

uaihebert:
> > i will change the approach then, and open a new PR.
> 
> Can you please provide more details before implementing? As I mentioned in #1110 it's not so obvious to me what this time counter is _supposed to display_. So, to be honest I'm not enthusiastic about changing the actual behavior. It would probably be useful to discuss what you want to do before implementing it (just to save time!)
> 
> And no, I don't think it's a good idea to move the time counter to the left toolbar. Reason for this is simple: the main toolbar is completely configurable, user can choose to display time counter or not (and where). Left toolbar is currently hard-coded. That would be more useful to make the left toolbar configurable also, but that's a significant job.

The idea is: i have an exercise that, if I push the "play" button it lasts for 20s. I want to repeat the same exercise for 5 minutes. Today I need another app to keep track of that, which makes harder to control tux + app + guitar + pick in hand. That is why a "global" counter (without being affected by speed) would come in handy.

Elapsed time is only for the current "loop", and as you described, it is affected by the speed. Then, I will change the approach and do it as a "command" of a separate component.

is it clear now? I totally understand about the left toolbar and agree with it.

guiv42:
> is it clear now?

Well, not really!
Let me try to reword, so you can tell me if I have understood. What you want is a "simple" timer, that progresses together with physical time (independently from tempo percentage), and that shows how long the player has been playing. If so, then the concept is slightly different from current `TGMainToolBarItemTimeCounter`.

So I can propose 2 options:
1. keep existing implementation unchanged, and create a new independent main toolbar item that implements the behavior you expect. Then, it may be useful to use a different style, since the 2 time counters may be displayed simultaneously in the toolbar.
2. Add an option to `TGMainToolBarItemTimeCounter`. Currently, when you click on the time counter it opens a font chooser dialog. You could replace this by a dedicated dialog, showing the selected font + a pair of radio buttons to choose the behavior (use song time reference / use wall clock). But definitely it's not a good solution to configure a toolbar item from another completely unrelated dialog (`TGTransportModeDialog` modified in your 2 PRs)

Option 2 may be better, as option 1 would probably get unnoticed by 99% of users who do not take the time to search in all available items to customize main toolbar.

Just an additional warning: in your 2 PRs you have made the method `MidiPlayer.isPaused()` public, clearly I don't recommend that.
Practically, the state of MidiPlayer is Boolean: it is playing or stopped. But it is *not obvious at all* to define what a "paused" player is. What's the difference between "paused" and "stopped"?
Current implementation is definitely not reliable. Example:
1. start playing
2. when in measure 4, hit "pause"
3. delete measure 4

What is the status of the player then? Logically it cannot be considered paused any more, because it can not restart from where it stopped. If measure 4 was the last one in the song, there is no more measure 4... Basically, any action editing the tab should provoke a transition "paused -> stopped", but currently it's not the case.
In your case it may be better to listen to play/pause/stop events to know when to start/stop/reset your counter.

guiv42:
@uaihebert, a completely independent comment: I've seen in your PRs that you have edited the Portuguese translation. If you have some knowledge in Portuguese, some available time, and if you want to contribute to TuxGuitar, that would really be nice if you could make a global update of the Portuguese translation.
It's not been modified since quite a long time, and we have added a significant number of features that require strings translations (just search for all lines starting with "# ")

uaihebert:
> > is it clear now?
> 
> Well, not really! Let me try to reword, so you can tell me if I have understood. What you want is a "simple" timer, that progresses together with physical time (independently from tempo percentage), and that shows how long the player has been playing. If so, then the concept is slightly different from current `TGMainToolBarItemTimeCounter`.
> 
> So I can propose 2 options:
> 
> 1. keep existing implementation unchanged, and create a new independent main toolbar item that implements the behavior you expect. Then, it may be useful to use a different style, since the 2 time counters may be displayed simultaneously in the toolbar.
> 2. Add an option to `TGMainToolBarItemTimeCounter`. Currently, when you click on the time counter it opens a font chooser dialog. You could replace this by a dedicated dialog, showing the selected font + a pair of radio buttons to choose the behavior (use song time reference / use wall clock). But definitely it's not a good solution to configure a toolbar item from another completely unrelated dialog (`TGTransportModeDialog` modified in your 2 PRs)
> 
> Option 2 may be better, as option 1 would probably get unnoticed by 99% of users who do not take the time to search in all available items to customize main toolbar.
> 
> Just an additional warning: in your 2 PRs you have made the method `MidiPlayer.isPaused()` public, clearly I don't recommend that. Practically, the state of MidiPlayer is Boolean: it is playing or stopped. But it is _not obvious at all_ to define what a "paused" player is. What's the difference between "paused" and "stopped"? Current implementation is definitely not reliable. Example:
> 
> 1. start playing
> 2. when in measure 4, hit "pause"
> 3. delete measure 4
> 
> What is the status of the player then? Logically it cannot be considered paused any more, because it can not restart from where it stopped. If measure 4 was the last one in the song, there is no more measure 4... Basically, any action editing the tab should provoke a transition "paused -> stopped", but currently it's not the case. In your case it may be better to listen to play/pause/stop events to know when to start/stop/reset your counter.

I did some changes, and now the player can add it as a command:
image
image

(the icon i will try to search change for a better icon, i just reused the tempo icon. i will search for a free one on the web)

As you can see, now it will not be affected by the tempo at 50% or so; it is a command, so the player will add it if he is willing to.

uaihebert:
> @uaihebert, a completely independent comment: I've seen in your PRs that you have edited the Portuguese translation. If you have some knowledge in Portuguese, some available time, and if you want to contribute to TuxGuitar, that would really be nice if you could make a global update of the Portuguese translation. It's not been modified since quite a long time, and we have added a significant number of features that require strings translations (just search for all lines starting with "# ")

sure, i can help with that. but i can help with portuguese from brazil, not pt_PT :D

guiv42:
OK, so you've gone for option 1 suggested above. Fine. Just take care about that `MidiPlayer.isPaused` issue.

For the translation, that would be a great opportunity to create the Brazilian version, there are currently many Brazilian users. Please tell me if you need support to create this new version.

uaihebert:
> OK, so you've gone for option 1 suggested above. Fine. Just take care about that `MidiPlayer.isPaused` issue.
> 
> For the translation, that would be a great opportunity to create the Brazilian version, there are currently many Brazilian users. Please tell me if you need support to create this new version.

About hte pause, i forgot to explain. If the user pauses and "unpauses" it, the session time should continue. That is the difference between pause and stop. if he pauses and removes something it should not affect what he is playing, the total session time remains the same. and when he pushes play, it will continue.

uaihebert:
> OK, so you've gone for option 1 suggested above. Fine. Just take care about that `MidiPlayer.isPaused` issue.
> 
> For the translation, that would be a great opportunity to create the Brazilian version, there are currently many Brazilian users. Please tell me if you need support to create this new version.

about the translation, would it be fine then if I create a new file based on the current one? and then just edit those # present there?

guiv42:
> if the user pauses and "unpauses" it, the session time should continue.

Yes, it was clear to me. This is why I suggest listening to play/pause/stop events instead of relying on MidiPlayer internal `isPaused()` method, that may not be reliable.

> about the translation, would it be fine then if I create a new file based on the current one? and then just edit those # present there?

Yes, that would be perfect!

uaihebert:
> > if the user pauses and "unpauses" it, the session time should continue.
> 
> Yes, it was clear to me. This is why I suggest listening to play/pause/stop events instead of relying on MidiPlayer internal `isPaused()` method, that may not be reliable.
> 
> > about the translation, would it be fine then if I create a new file based on the current one? and then just edit those # present there?
> 
> Yes, that would be perfect!

i will do the translation first, later on i will try to do the current approach with the event. i understood what you mean now.

PR #1114, opened 2026-07-10
add session time using event listener approach

uaihebert:
Add session count timer.
To use it:
image
When the player press play with looped option, we will see the total time:
image

Icon is free to use, got it from here: https://fontawesome.com/icons/classic/solid/stopwatch

### DIFF
--- a/common/TuxGuitar-lib/src/main/java/app/tuxguitar/player/base/MidiPlayer.java
+++ b/common/TuxGuitar-lib/src/main/java/app/tuxguitar/player/base/MidiPlayer.java
@@ -133,6 +133,7 @@ public class MidiPlayer{
 			this.setPaused(paused);
 			this.stopSequencer();
 			this.setRunning(false);
+			this.notifyStopped(paused);
 		} finally {
 			this.unlock();
 		}
@@ -164,6 +165,7 @@ public class MidiPlayer{
 
 			final boolean notifyStarted = (!this.isRunning());
 			this.setRunning(true);
+			this.setPaused(false);
 			this.stopSequencer();
 			this.checkDevices();
 			this.updateLoop(true);
@@ -307,7 +309,7 @@ public class MidiPlayer{
 				}
 
 				if(!this.isRunning() ){
-					this.notifyStopped();
+					this.notifyStopped(this.isPaused());
 				}
 			} finally {
 				this.unlock();
@@ -1261,8 +1263,8 @@ public class MidiPlayer{
 		TGEventManager.getInstance(this.context).fireEvent(new MidiPlayerEvent(MidiPlayerEvent.NOTIFY_STARTED));
 	}
 
-	public void notifyStopped(){
-		TGEventManager.getInstance(this.context).fireEvent(new MidiPlayerEvent(MidiPlayerEvent.NOTIFY_STOPPED));
+	public void notifyStopped(boolean paused){
+		TGEventManager.getInstance(this.context).fireEvent(new MidiPlayerEvent(MidiPlayerEvent.NOTIFY_STOPPED, paused));
 	}
 
 	public void notifyCountDownStarted(){
--- a/common/resources/lang/messages.properties
+++ b/common/resources/lang/messages.properties
@@ -290,6 +290,7 @@ toolbar.settings.save-as=Save current toolbar as
 toolbar.settings.delete=Delete current toolbar
 toolbar.separator=-------- Separator --------
 toolbar.timeCounter=Elapsed time
+toolbar.playingSessionTime=Playing session time
 toolbar.tempoIndicator=Tempo indicator
 toolbar.main.left=Left area
 toolbar.main.center=Center area
--- a/common/resources/lang/messages_pt.properties
+++ b/common/resources/lang/messages_pt.properties
@@ -147,6 +147,7 @@ toolbar.settings.save-as=Salvar Como
 toolbar.settings.delete=Excluir
 # toolbar.separator=
 toolbar.timeCounter=Tempo Decorrido
+toolbar.playingSessionTime=Tempo de prática
 toolbar.tempoIndicator=Indicador de Tempo
 toolbar.main.left=Área Esquerda
 toolbar.main.center=Área Central
--- a/desktop/TuxGuitar/src/app/tuxguitar/app/view/toolbar/main/TGMainToolBarSectionSessionTime.java
+++ b/desktop/TuxGuitar/src/app/tuxguitar/app/view/toolbar/main/TGMainToolBarSectionSessionTime.java
@@ -1,0 +1,154 @@
+package app.tuxguitar.app.view.toolbar.main;
+
+import app.tuxguitar.app.TuxGuitar;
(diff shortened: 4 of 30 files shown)


common/resources/lang/messages.properties
```diff
@@ -290,6 +290,7 @@ toolbar.settings.save-as=Save current toolbar as
 toolbar.settings.delete=Delete current toolbar
 toolbar.separator=-------- Separator --------
 toolbar.timeCounter=Elapsed time
+toolbar.sessionTime=Current Session Time
```
guiv42: The word "session" is not really clear to me.
Proposal, replace by `toolbar.playingSessionTime=Playing session time"`

common/resources/lang/messages_pt.properties
```diff
@@ -147,6 +147,7 @@ toolbar.settings.save-as=Salvar Como
 toolbar.settings.delete=Excluir
 # toolbar.separator=
 toolbar.timeCounter=Tempo Decorrido
+toolbar.sessionTime=Sessão Atual
```
guiv42: can you please add explicitly in all languages the new string to be translated, with a "#" prefix?
Else, contributors may not identify that something new needs to be translated

desktop/TuxGuitar/share/skins/Breeze-dark/session_time.png
guiv42, line 1: That's not the right way to create a new icon.
You need first to create a vector graphics version (svg), and then the png (light/dark) can be derived from it.
See https://github.com/helge17/tuxguitar/tree/master/desktop/TuxGuitar/src-skins

possible simple alternative: remove the icon

common/TuxGuitar-lib/src/main/java/app/tuxguitar/player/base/MidiPlayer.java
guiv42, line 1: I'm still not very comfortable with this "paused" state.
Need to spend more time on that question

desktop/TuxGuitar/src/app/tuxguitar/app/view/toolbar/main/TGMainToolBarSectionSessionTime.java
guiv42, line 1: This requires some fine-tuning about the font.
I'm not quite sure about which default font will be used on each system (win/mac/linux/bsd), but in my configuration it is not very readable (too small).
Another problem: I highly recommend to use a fixed-width font. If not, user may see some horizontal movement whenever the length of the text string changes. Furthermore, the size of the control may be incorrect in some cases.
You can look for inspiration in the existing time counter, that has different fonts set up for each system (plus an option to set font manually, that may be useful if there's an issue with default config)

guiv42 (commented):
First very short review (I'll need to find some more time to look at the midi player events management)

One open question about the "timer" icon: is it really necessary? I don't think so. If it's removed, then this timer could probably become a standard `TGMainToolBarItem`, with no need to create a new toolbar item type. That would be much simpler to maintain.

uaihebert:
ty, tomorrow I will look into your comments. ( :

uaihebert:
> First very short review (I'll need to find some more time to look at the midi player events management)
> 
> One open question about the "timer" icon: is it really necessary? I don't think so. If it's removed, then this timer could probably become a standard `TGMainToolBarItem`, with no need to create a new toolbar item type. That would be much simpler to maintain.

one question about this:
image
I liked the idea. I will remove the icon and put it here (if you agree to it) as a fixed standard item. 

Is it fine like that?

common/TuxGuitar-lib/src/main/java/app/tuxguitar/player/base/MidiPlayer.java
uaihebert: sure, np

uaihebert:
@guiv42 i pushed the changes, and the code is indeed simpler. 
image
this is how it looks like, and it is using the same font as the elapsed time component.
image
adding a new icon works as expected

guiv42 (changes requested):
> I will remove the icon and put it here (if you agree to it) as a fixed standard item.

No, I clearly disagree.

There is currently one single fixed item in the toolbar, it's the toolbar settings button in the last position of the right part of the toolbar. *Everything else can be deleted by user*. This new feature definitely does not deserve to be imposed to everyone: it cannot be fixed.

Another important thing: this very last position in the right part of the toolbar has a specific behavior. Basically, this place is always dedicated to the "toolbar settings" button, and using the UI it's impossible to change that.
That is pretty important, as if this button is removed (e.g. accidentally) there is no more any possibility to modify the toolbar.


I did not take time to review completely your last implementation, I just checked how you handled the font issue, and that's not correct. You are using a parameter that the user can define when he/she clicks on the already existing timestamp indicator. Said differently: to change the font of your new indicator, you need to click on the other one (that may not even be present in the toolbar).

In order to avoid a significant complexity, let me come back to a [proposal](https://github.com/helge17/tuxguitar/pull/1111#issuecomment-4924569647) I have made in your previous PR (I named it "option 2"):
Instead of creating a new item for the toolbar, you could re-use the existing time counter and let the user configure what he/she wants to see. Either a *song timestamp*, or a *playing session timestamp*.

This way, you could take benefit of the existing implementation that already handles some complexity.
The idea would be to change the behavior when user clicks on the timestamp: now it opens a font chooser dialog, it could open a new "timestamp configuration" dialog. This dialog would be pretty simple: a button to launch the font chooser, plus 2 radio buttons to choose the behavior.

uaihebert:
> > I will remove the icon and put it here (if you agree to it) as a fixed standard item.
> 
> No, I clearly disagree.
> 
> There is currently one single fixed item in the toolbar, it's the toolbar settings button in the last position of the right part of the toolbar. _Everything else can be deleted by user_. This new feature definitely does not deserve to be imposed to everyone: it cannot be fixed.
> 
> Another important thing: this very last position in the right part of the toolbar has a specific behavior. Basically, this place is always dedicated to the "toolbar settings" button, and using the UI it's impossible to change that. That is pretty important, as if this button is removed (e.g. accidentally) there is no more any possibility to modify the toolbar.
> 
> I did not take time to review completely your last implementation, I just checked how you handled the font issue, and that's not correct. You are using a parameter that the user can define when he/she clicks on the already existing timestamp indicator. Said differently: to change the font of your new indicator, you need to click on the other one (that may not even be present in the toolbar).
> 
> In order to avoid a significant complexity, let me come back to a [proposal](https://github.com/helge17/tuxguitar/pull/1111#issuecomment-4924569647) I have made in your previous PR (I named it "option 2"): Instead of creating a new item for the toolbar, you could re-use the existing time counter and let the user configure what he/she wants to see. Either a _song timestamp_, or a _playing session timestamp_.
> 
> This way, you could take benefit of the existing implementation that already handles some complexity. The idea would be to change the behavior when user clicks on the timestamp: now it opens a font chooser dialog, it could open a new "timestamp configuration" dialog. This dialog would be pretty simple: a button to launch the font chooser, plus 2 radio buttons to choose the behavior.

Hello again, i guess I misunderstood you then.

I corrected the approach here: https://github.com/helge17/tuxguitar/pull/1119

PR #1129, opened 2026-07-22
feat: tonic note highlighting on fretboard

yaadyaad:
- New config key: fretboard.color.tonic (default 0,0,0 — same as fretboard)
- Tonic note in selected scale rendered with distinct color
- 'Tonic Note Color' picker in Fretboard Settings dialog
- Applied to both oval and text rendering modes

image
image

### DIFF
--- a/common/resources/lang/messages_bg.properties
+++ b/common/resources/lang/messages_bg.properties
@@ -161,6 +161,7 @@ fretboard.string-color=Цвят на струна
 fretboard.fretpoint-color=Цвят на прагче
 fretboard.note-color=Цвят на нота
 # fretboard.scale-note-color=
+# fretboard.tonic-color=
 fretboard.direction=Посока
 fretboard.right-mode=Десен режим
 fretboard.left-mode=Ляв режим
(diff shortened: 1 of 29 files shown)


common/resources/lang/messages_bg.properties
```diff
@@ -161,6 +161,7 @@ fretboard.string-color=Цвят на струна
 fretboard.fretpoint-color=Цвят на прагче
 fretboard.note-color=Цвят на нота
 # fretboard.scale-note-color=
+fretboard.tonic-color=Tonic Note Color
```
guiv42: That's not the correct way to update translations.
- for languages you know, put the explicit translation (as you did in `messages.properties` for the English version)
- for all languages you *don't* know, do not translate the string and add a "# " prefix. Here:
`# fretboard.tonic-color=`

This way, contributors can easily identify what needs to be translated.

guiv42 (changes requested):
Just one remark about translation files, else it's OK
thanks for the contribution

common/resources/lang/messages_bg.properties
yaadyaad: Thank you for reviewing, I have applied the fix.

guiv42 (approved):
OK

PR #1130, opened 2026-07-22
feat: Open Recent menu and auto-load last file on startup

yaadyaad:
- Added submenu File -> Open Recent, right after Open URL
- Auto-load last opened file on startup (config key: auto.load-last-file, default: true)
- Uses existing TGFileHistory infrastructure for tracking recent files

image

### DIFF
--- a/common/resources/lang/messages.properties
+++ b/common/resources/lang/messages.properties
@@ -186,6 +186,7 @@ file.new-song.default-template=New Song
 file.open=Open
 file.open.error=Cannot Open: {0}
 file.open-url=Open URL
+file.open-recent=Open Recent
 file.custom-template.select=Set custom template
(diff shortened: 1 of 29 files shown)


guiv42 (commented):
Sorry but I don't understand this PR.

From what I see:
1. it moves one item in the file menu
2. it updates one English translation
3. it provides an option to reload the (?) last open file on startup

For 1. and 2., I don't see any reason to do this. This menu item has been there for years and people are used to it. By the way, the implementation breaks all non-English translations, and leaves one unused item in non-English translation files.

For 3., it's not so clear as several files may be open when the app is closed. By the way, a more complete implementation of this feature is already in progress (see PR #1044).

So I propose to close this PR as is.

PR #1131, opened 2026-07-22
feat: Show Note Names on the Score Staff                                                   

yaadyaad:
Show Note Names on the Score Staff                                                   
                                                                                                      
 A "learning mode" feature that displays pitch names (C, D#, Eb, F, etc.) next to each note on the    
 standard musical staff. This helps beginners learn to read sheet music by showing what each note is  
 called.                                                                                              
                                                                                                      
 How it works:                                                                                        
 - When enabled, each note on the score staff gets a text label with its pitch name (e.g. "C", "D#",  
   "Eb")                                                                                              
 - Labels appear just to the left of the note head, aligned with its vertical position                
 - Pitch names respect the current key signature — correct accidentals (sharps/flats) are displayed   
 - Uses a stable default font (no visual jumping)                                                     
 - Only applies to the score view, not tablature or percussion tracks                                 
                                                                                                      
 How to toggle:                                                                                       
 - View → Show Note Names (menu item)                                                                 
 - A toolbar button is also available (can be added via Toolbar Settings)                             
                                                                                                      
 Changes:                                                                                             
 - New layout flag DISPLAY_NOTE_NAME (0x0200) in TGLayout                                             
 - TGNoteImpl.paintScoreNote() — renders pitch name text per note                                     
 - TGSetNoteNameEnabledAction — toggles the flag                                                      
 - Menu + toolbar registration                                                                        

image

image

### DIFF
--- a/common/TuxGuitar-lib/src/main/java/app/tuxguitar/graphics/control/TGLayout.java
+++ b/common/TuxGuitar-lib/src/main/java/app/tuxguitar/graphics/control/TGLayout.java
@@ -32,6 +32,7 @@ public abstract class TGLayout {
 	public static final int DISPLAY_MODE_BLACK_WHITE = 0x40;
 	public static final int HIGHLIGHT_PLAYED_BEAT = 0x80;
 	public static final int CONTINUOUS_SCROLL = 0x0100;
+	public static final int DISPLAY_NOTE_NAME = 0x0200;
 
 	private int style;
 	private float scale;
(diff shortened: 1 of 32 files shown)


guiv42 (changes requested):
I'm definitely **not** in favor of this feature, it has a too high potential for graphical conflicts. See screenshot below, after 2 minutes of testing (not reviewed the code):



(by the way, last note name is incorrect, but that's not the point).

Adding this kind of information requires to review the horizontal spacing in the measure. And that would be pretty tricky, especially not to break alignment of scores in multi-track configuration.

Proposal to close PR.

Issue #1146, opened 2026-08-01
play repeats in loop mode


belfie13:
currently if you loop between frets and the last fret is a repeat it will stop playing before it repeats.

the only way to repeat it is to select the next measure as well. so you can't play the bar with repeats.

to reproduce:
- create three bars
- put a repeat on the first bar and the second bar.
- select the first two bars
- play in a loop

the first bar plays once then repeats but the second bar only plays once.

```
         x1         x1
|:========:|:========:|==========|
|<-------select------>|             and play looped
|<plays 2x>|<plays 1x>|<plays 0x>|

should:
|<plays 2x>|<plays 2x>|<plays 0x>|
```

guiv42:
Mixing loop and repeats is a pretty complex scenario in terms of implementation.
Already fixed a few bugs in some similar configurations, and in some cases it's even not so obvious to decide what is the best behavior (typ. with loops starting in the middle of alternative repeats)...

Sumit2342:
Hi @guiv42 ,
I was checking this issue and found it is actually a side effect of the fix for [#549.](https://github.com/helge17/tuxguitar/issues/549 ) To stop the player from freezing when a loop ends on an alternative repeat, a condition was added in **MidiRepeatController.java** that tells the parser to completely ignore Repeat Close commands if they land on the last measure of a loop.
I tested removing that strict eHeader check and it fixes this issue. It doesn't affect the changes for #549. 

I have a fix for this, [PR](https://github.com/helge17/tuxguitar/pull/1147/commits)

guiv42:
fixed in master branch by #1147 thanks to @Sumit2342

Issue #549, opened 2024-10-11
Bug: Failure to start playing when start of loop range coincides with non-first "alternative repeat"


lucmans:
When starting playback of tabs, with the loop range starting on an alternative repeat that is not the first alternative, playback stalls.
See the added screenshot for reference.
![2024-10-11-185056_252x151_scrot](https://github.com/user-attachments/assets/e1ad1869-2582-4b4b-9384-018d11ab9866)

guiv42:
reproduced with TuxGuitar 1.6.4

guiv42:
Very similar to #234 , but it seems the conditions to reproduce are not exactly the same.
See video below, in this case it works fine even for alternative repeat 4.

https://github.com/user-attachments/assets/6a14f43c-e3ae-48e7-b282-8a3fbf40767a



@lucmans : could you please share what is the repeat structure before your measure 18?

lucmans:
[tab.zip](https://github.com/user-attachments/files/17351410/tab.zip)
I uploaded the file for reference.

guiv42:
Understood
When parsing song, MidiRepeatController [ignores](https://github.com/helge17/tuxguitar/blob/c841f2391a38d4ee5a9890c2be28b2219046e780/common/TuxGuitar-lib/src/org/herac/tuxguitar/player/base/MidiRepeatController.java#L49) every measure out of the defined loop.
A first fix was applied (94bbab607cf7d0835eb8430534542b9570d23350) to [consider simple repeats](https://github.com/helge17/tuxguitar/blob/c841f2391a38d4ee5a9890c2be28b2219046e780/common/TuxGuitar-lib/src/org/herac/tuxguitar/player/base/MidiRepeatController.java#L48), but it's not sufficient.
If a measure belongs to several alternative repeats, the following "repeat close" shall be executed several times. Currently it is only considered once. See provided example file, measure 17 belongs to alternative repeats 1,2, and 3. This results in an incorrect computation of what measures should be played in the loop.
Fix coming soon.

guiv42:
should be better now

guiv42:
Fix is available from pre-release [2024-10-23](https://github.com/helge17/tuxguitar/releases/tag/2024-10-23-master), **in master branch**

PR #1147, opened 2026-08-02
Fix repeat logic in MidiRepeatController and update tests for loop sequence

Sumit2342:
Fix repeat logic in MidiRepeatController and update tests for loop sequence


guiv42:
see #1146

Thanks for the PR. I need to spend some time to analyze: this PR does not only change the implementation, but also the expected behavior in the test case. Especially, I need to remember why I have implemented it this way. See the comment above the modified line, ignoring the repeat close in this case was purely intentional.
By the way: if the modification implemented by this PR is to be merged, then this comment needs to be modified also.

Sumit2342:
Thanks for taking a look

The reason this PR changes the test expectations is that we actually don't need to truncate the repeat to prevent #549 freeze anymore. By allowing the repeat to process fully (which changes the expected output arrays from {1,2,3} to {1,2,3,1} and {4,5} to {4,5,4,5}), the sequence resolves naturally. It finishes without getting stuck because the background state calculation successfully catches the bounds and stops the sequence on its own. 

So, we get to keep the safety against infinite loops, while also ensuring that self-contained repeats │ (like looping a single measure with a repeat) play fully instead of dropping a pass.

Also, I missed to update the comment, I have updated it now.

guiv42:
It's not that obvious. Refer to the existing test case, with this PR it is no more possible to play measures {1,2,3} in a loop, which seems a reasonable use case for me:
test

A possible alternative, if the last measure of the loop contains a "repeat close":
only consider the repeat close if the corresponding repeat open is inside the loop, but *strictly after* the first measure of loop.

A bit more complex to implement, but that would leave the test case unchanged (that makes sense for me), and also fix this bug.

@Sumit2342 what do you think?

Sumit2342:
yes you are correct, thanks for clarification
Case1:
|:========|========|========:|
|<----------select---------->| and play looped (M1 to M3)

here the loop starts from M1 to M3 and repeat open exactly at M1 not after M1. So we should ignore the repeat and the music will be played as 1>2>3>1>2>3.....

Case2:
|=========|:=======|========|========:|
|<--------------select----------------------->| and play looped (M1 to M4)

here the loop starts from M1 to M4 and repeat open after M1, which is strictly after loop starts. So we should not ignore the repeat. So, it should play as 1>2>3>4>2>3>4>1>2>3>4>2>3>4....

This should be the correct behavior right @guiv42 ?

Sumit2342:
I have updated the PR, please let me know if it is as per expected behavior.

guiv42 (approved):
Yes, that looks good, thanks.

Could you please just add a test case corresponding to #1146?
Tell me if you prefer me to do it.

Sumit2342:
Hi,
I have added the test case . Please review it and let me know if this looks good to you.

guiv42:
I should have been more precise in my request for the new test case.
Instead of creating a new .tg file, I think it would be more simple (and consistent) to add a few measures with the right repeats to the existing "altRepeatLoop_20.tg" file

Sumit2342:
I removed the new .tg file and added 3 new measures. I also updated the testcase accordingly. Please review it again let me know if this looks good.

guiv42:
That's fine, thanks a lot!

Issue #1020, opened 2026-02-26
Feature request: Add more shortcuts for moving the cursor


wjandrea:
I would like:
Home = move cursor to start
End = move cursor to end

Also could be useful:
Page Up = move cursor one page up (or one page left in multitrack mode)
Page Down = move cursor one page down (or one page right in multitrack mode)

guiv42:
go to start / end should be quite simple to implement.
"page" up or down is significantly more complex

LeonardoQF:
I think those shortcuts already exist:
Go to first measure: Ctrl + Shift + Left
Go to last measure: Ctrl + Shift + Right

Both move the cursor (Caret) to their respective positions as well.

woody6402:
See discussions page up/down, where a patch is under preparation

woody6402:
> I would like: Home = move cursor to start End = move cursor to end 
> Also could be useful: Page Up = move cursor one page up (or one page left in multitrack mode) Page Down = move cursor one page down (or one page right in multitrack mode)

One questions since we do that discuss on another page. When raising the requestr did you have a movement of the caret (edit cursor) or moevement of the view in mind.

Issue #655, opened 2025-02-23
some shortcuts are not displayed in menus


guiv42:
TuxGuitar 1.6.6, Linux (openSuse tumbleweed)
problem present since 1.6.0, it was OK with 1.5.6.

Some shortcuts are not displayed in menu items. Example (not exhaustive list), in menu "Beat":
- not displayed:
  - Insert Rest Beat: "Ins"
  - Delete Note or Rest: "Del"
  - Clean Beat: "Ctrl+Del"
- correctly displayed:
  - Insert Text: "T"

It looks like another SWT issue (similar to #33), probably linked to the representation of some specific keys (Del, Ins)

guiv42:
Good news, analysis of the SWT issue is progressing. There is currently an open PR that would fix the issue (probably not in next release, maybe in the following one)

guiv42:
still good news, the [PR](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2173) I mentioned above has been merged. So we can expect a solution to this issue with next release of SWT (4.37)

guiv42:
Confirmed, issue is fixed in TuxGuitar with SWT 4.37

guiv42:
SWT is now updated to 4.37 in tuxguitar-next branch. Fix will be available in next minor release

PR #1180, opened 2026-09-02
Add Page Up and Page Down view scrolling

woody6402:
Adds Page Up and Page Down actions for scrolling the tablature vertically by approximately 80% of the visible page.
  The actions are registered with default keyboard shortcuts and work on the currently selected tab.

  Special handling is required because caret.hasChanges() may become true during redraw even when the caret position has
  not actually changed. Without this handling, TuxGuitar immediately jumps back to the caret after a manual page scroll.

  The implementation therefore records the caret’s track, position, and string when manual scrolling begins. It only
  resumes automatic caret-following when one of those values genuinely changes; otherwise, the spurious hasChanges()
  flag is cleared and the manually selected scroll position is preserved.


woody6402:
Needs oncemore a look if there's a conflict with playmode and scrollig.

=> Yes there's an interference with play mode, which is solved. After running some tests, I will  reopen ...

common/resources/lang/messages.properties
```diff
@@ -778,6 +778,8 @@ action.edit.toggle-free-edition-mode=Toggle Free Edition Mode
 action.edit.display-errors=Display measure errors
 action.caret.go-up=Move cursor up
 action.caret.go-down=Move cursor down
+action.view.scroll-page-up=Scroll view one page up
```
guiv42: When adding new strings for translation, please add the same 2 lines with a "#" prefix in all other languages.
This makes explicit to contributors that something needs to be translated

desktop/TuxGuitar/src/app/tuxguitar/app/view/component/tab/TGControl.java
guiv42, line 1: I've not yet started a deep review of the implementation, I just started PR analysis with a few tests.

**First test:**
- open a song, display one single track in page layout (choose a song that fills more that 1 page)
- make sure caret is in first measure of track
- hit "PgDn" key:
  - display is scrolled by one page as expected
  - but the caret is invisible. Hit "right arrow" key and the focus comes back to measure 1
 
Implementing "PgUp/PgDn" features in `TGControl` seems counter-intuitive: it scrolls the display but does not move the caret.
It would seem more logical to move the caret, rather than just scrolling the displayed area. It may be a better solution to elaborate a target position for the caret, then to call `TGMoveToAction`. This should automatically trigger an update of the displayed area.
What do you think?

**Second test:**
- still in page mode, single track displayed
- play song
- when currently played measure is about the middle of the screen, hit "PgUp" or "PgDn" key
- observed behavior: nothing happens (same behavior with and without "continuous scrolling")

Personally I'm fine with that. Automatic scrolling, be it discrete or continuous, is a rather complex feature that already showed several bugs in the past. Interference between automatic scrolling and manual actions (scrollbars, tracks table, etc.) is already a pain, so I like the idea to keep it simple.
But in this case, it would be much safer to inhibit the feature while playing from `TGActionConfigMap`, by setting `DISABLE_ON_PLAY` flag. This way the action would not be triggered at all when shortcut keys are hit while playing. Much cleaner than to launch an action and to check player status while processing the action.

common/resources/lang/messages.properties
woody6402: Yes, I will do ...

desktop/TuxGuitar/src/app/tuxguitar/app/view/component/tab/TGControl.java
woody6402: **First test:**

This was a deliberate design decision. For example, I may be editing near the end of the score, scroll up to look at
 an earlier section, and then want to continue editing at the original caret position.

However, I agree that there may be other use cases where moving the caret together with the displayed page would be preferable. I am not yet sure which behavior would be the best overall solution.

**Second test:**

I will look into using DISABLE_ON_PLAY, since this seems like the more natural solution and should replace my initial check inside the action.

During my tests, the interaction between playback scrolling and Page Up/Page Down—especially within repeat sections— became very confusing. I therefore deliberately decided to disable these actions during playback.

I will check the translation entries and the DISABLE_ON_PLAY handling and provide a follow-up patch. Regarding moving the caret, I would like to consider the expected behavior a little further before changing it.

guiv42: > Regarding moving the caret, I would like to consider the expected behavior a little further before changing it.

A few arguments:
- in TuxGuitar, the 4 arrows make the *caret (or cursor)* move
- just open any text editor (or spreadsheet, or ...), and hit "PgDn" key: the *cursor* moves down by 1 page
- refer to title of feature request #1020

Take care, I do not consider this as a simple feature. I've had quite a few issues in the past just trying to move the caret. Especially in this case, the behavior should take into consideration the current layout to dertermine a target beat for the caret (e.g. in *linear mode* the concept of "page" is unclear).

woody6402: Yes that's far more complicated.

But maybe we have to consider two different user groups:
- people who view and play the piece (80%), which want to scroll down the page (without having the caret in mind)
- people who edit. They may want to move the caret. Which leads to more questions like:
-- page layout  => maybe caret follows view (or better from implementation side, view follows caret)  
-- multitrack => ?
-- linear mode => ? (as you have mentioned above)

To be honest, I'm coming mainly from the first and by using tuxguitar for 20 years I was always heavily missing the move one side down button/key when startig to play without training mode.

Also took a look on Move to the End/Beginning => there caret is moved and the tab/guitar side is preserved. I think I need to look inside this part.

guiv42: It's not only about "view vs. edit". Whenever you scroll down one page and then you hit "play", I find it very confusing if it starts playing from some place that is not visible.

Anyway, since these actions are only triggered by keyboard shortcuts, we could as well consider the 2 use cases, and have different actions co-existing: "scroll to" and "move to".
In this case, I would prefer to have the default keys PgUp/PgDn associated to "move to" actions, just like any other editor.

woody6402: I have pushed a new version in , which I need to test. 

But the first test succeeded for multitrack and single track including moving the caret and reusing the existing positioning in the same way  as move to the end/beginning.

But it currently switches to line up/down.

I will give an OK for further review when the tests are ready.

Mainly:
- The implementation starts by retrieving the current caret position: track, measure, beat, string, voice, and the active layout.
- It then searches within the current track for the previous or next distinct vertical measure position. Using the
measure’s posY rather than the global layout-line number is important for multitrack mode, where consecutive layout- line numbers may refer to different tracks within the same system.
- After finding the adjacent line, the current beat’s horizontal position is used to select the corresponding measure on that line. Within the selected measure, the relative musical position of the original beat is calculated and mapped to the target measure. The closest available beat in the current voice is preferred, with a fallback to any available beat. This ensures, for example, that the first beat maps to the first beat even when line indentation, clefs, or time signatures produce different graphical offsets.
- Finally, the target track, measure, beat, and current string are passed to the existing TGMoveToAction. The standard caret-change and display-follow mechanisms then bring the target into view, without directly manipulating the scrollbar. Playback is excluded centrally through DISABLE_ON_PLAY.

woody6402:
Preview:
https://github.com/user-attachments/assets/6e844979-8960-4842-959a-3451a560cbe7

guiv42:
I'm sorry, but I'm getting lost, I don't really understand:
```
action.view.scroll-line-up=Move caret one score line up
action.view.scroll-line-down=Move caret one score line down
action.view.scroll-page-up=Scroll view one page up
action.view.scroll-page-down=Scroll view one page down
```

I was expecting something like "*move caret* 1 page" and "*scroll view* 1 page".
Why are you referring to "lines"?
Furthermore, if some actions scroll and others move the caret, then they should probably be renamed. To be more explicit, I was expecting something like this:
```
action.view.move-page-up=Move caret one page up
action.view.scroll-page-up=Scroll view one page up
(same for "down")
```

By the way, I'm really not comfortable about retrieving the geometrical position of graphical items from the action layer. Importing classes like `*Impl.java` feels like breaking some kind of separation between logic and graphical representation. I've got to spend more time on this, let's see if I can come up with something simpler (not guaranteed!)
But I first need to understand this "page vs. line" open point.

woody6402:
> I'm sorry, but I'm getting lost, I don't really understand:
> 
> ```
> action.view.scroll-line-up=Move caret one score line up
> action.view.scroll-line-down=Move caret one score line down
> action.view.scroll-page-up=Scroll view one page up
> action.view.scroll-page-down=Scroll view one page down
> ```
> 
> I was expecting something like "_move caret_ 1 page" and "_scroll view_ 1 page". Why are you referring to "lines"? Furthermore, if some actions scroll and others move the caret, then they should probably be renamed. To be more explicit, I was expecting something like this:
> 
> ```
> action.view.move-page-up=Move caret one page up
> action.view.scroll-page-up=Scroll view one page up
> (same for "down")
> ```

since I have switched to move the caret and as result all 4 cases are moving the caret.
The line movement is simple a side result from the page movement.  
At the end when already moving the caret there's no need to have a different key which moves the page without caret.

**And I see, I have misnamed it. All 4 keys are moving the caret it has to be** (will use cursor in english in the futire)
```

action.view.scroll-line-up=Move caret one score line up
action.view.scroll-line-down=Move caret one score line down
action.view.scroll-page-up=Move caret one page up
action.view.scroll-page-down=Move caret one page down
```

As said, it's not ready yet, currently there's a result which moves the caret, maybe not the final one, but it does the job for page layout in  simple and multitrack mode.

 
> By the way, I'm really not comfortable about retrieving the geometrical position of graphical items from the action layer. Importing classes like `*Impl.java` feels like breaking some kind of separation between logic and graphical representation. I've got to spend more time on this, let's see if I can come up with something simpler (not guaranteed!) But I first need to understand this "page vs. line" open point.

maybe needs a further look inside to beware te separation

woody6402:
I looked at how the existing caret navigation is structured. Actions such as TGGoLeftAction, TGGoRightAction,
  TGGoUpAction, and TGGoDownAction delegate the actual movement to methods on Caret.

  Following the same pattern, a cleaner implementation could be:

  Caret.moveLineUp()
  Caret.moveLineDown()
  Caret.movePageUp(float visibleHeight)
  Caret.movePageDown(float visibleHeight)

  All four methods could share one private implementation in Caret. The line methods would select the adjacent score
  line, while the page methods would receive the visible height from TGControl.

  The corresponding actions would then only trigger the operation:

  TGGoLineUpAction   -> Caret.moveLineUp()
  TGGoLineDownAction -> Caret.moveLineDown()
  TGGoPageUpAction   -> Caret.movePageUp(control.getVisibleHeight())
  TGGoPageDownAction -> Caret.movePageDown(control.getVisibleHeight())

  TGControl would remain responsible for the viewport height and the one-shot scroll adjustment associated with page
  movement. The geometrical interpretation of posX and posY, as well as the use of the graphical *Impl classes, would
  move into Caret, where these classes are already used by the existing navigation methods.

  This would remove graphical implementation details from the action package, keep the four actions small, and follow
  the established structure of the current caret navigation. The action names and IDs could consequently become
  TGGoLineUp/DownAction and TGGoPageUp/DownAction, using action.caret.go-* IDs rather than action.view.scroll-*.

guiv42:
> it's not ready yet

OK, take your time. Could you just explicitly request a review when you're ready?

> All four methods could share one private implementation in Caret

Yes, that looks a cleaner implementation to me. At least well aligned with current design.

woody6402:
> > All four methods could share one private implementation in Caret
> 
> Yes, that looks a cleaner implementation to me. At least well aligned with current design.

Then I will move the implementation to Caret and raise my hand when it's ready.

woody6402:
Move to caret done (see above) and tests are running.

guiv42:
OK, thanks for the update, I'll have a look.

common/resources/lang/messages.properties
```diff
@@ -778,6 +778,10 @@ action.edit.toggle-free-edition-mode=Toggle Free Edition Mode
 action.edit.display-errors=Display measure errors
 action.caret.go-up=Move cursor up
 action.caret.go-down=Move cursor down
+action.caret.go-line-up=Move cursor one score line up
```
guiv42, line 781: thank you for renaming the actions, it's much clearer now

desktop/TuxGuitar/src/app/tuxguitar/app/view/component/tab/TGControl.java
guiv42, line 1: I'm still really not favorable to the direct interaction between "move actions" and scrolling logic in `TGControl`, for several reasons:
- dynamic behavior of scrolling in TGControl is tricky, and it took time to stabilize it. With this implementation I'm not completely confident there will not be some unexpected side effects.
- for homogeneity with similar actions. Refer to already implemented "move to" actions (left/right arrow, alt+left/right arrow, ...): all these actions effectively move the cursor, but do not trigger any scrolling except when it's required to make the new cursor position visible.
  Your call to `moveTo` in `Caret.moveVertical` implementation already sets `Caret.changes` to `True`. That already triggers scrolling by `TGControl` *if needed* (and I can't exclude a possible race condition with your direct calls from `Caret` to `TGControl`)

So, please revert the modifications of `TGControl`, and remove the corresponding calls.

desktop/TuxGuitar/src/app/tuxguitar/app/view/component/tab/Caret.java
```diff
@@ -259,6 +260,131 @@ public void moveLeft() {
...
+		return this.moveVertical(1, visibleHeight);
+	}
+
+	/**
+	 * Moves between score lines while preserving the approximate horizontal and
+	 * musical position. Measures sharing the same {@code posY} belong to the same
+	 * score line. Searching only the selected track prevents a multi-track layout
+	 * from redirecting the caret to another track.
+	 */
+	private boolean moveVertical(int direction, float verticalDistance) {
```
guiv42: name of second parameter is somehow ambiguous: what does "moving vertically" mean if `verticalDistance == 0`?
My remark is just for clarity/readability of the code, not from functional point of view.

Suggestion to rename parameter. Would `minVerticalDistance` be better? Or maybe `pageVerticalDistance`? (at least that would make explicit the difference between *line* and *page*)

desktop/TuxGuitar/src/app/tuxguitar/app/action/impl/caret/TGGoPageDownAction.java
```diff
@@ -0,0 +1,36 @@
...
+public class TGGoPageDownAction extends TGActionBase {
+
+	public static final String NAME = "action.caret.go-page-down";
+
+	public TGGoPageDownAction(TGContext context) {
+		super(context, NAME);
+	}
+
+	protected void processAction(TGActionContext context) {
+		TGControl control = TGTabFolder.getInstance(getContext()).findSelectedControl();
```
guiv42: It seems more logical to leave `Caret` evaluate the amplitude of the required movement.
Something like `caret.movePageDown()` (without argument) should probably be sufficient.

desktop/TuxGuitar/src/app/tuxguitar/app/action/impl/caret/TGGoPageDownAction.java
```diff
@@ -0,0 +1,36 @@
...
+		if (control != null && !control.isDisposed()) {
+			Tablature tablature = TablatureEditor.getInstance(getContext()).getTablature();
+			Caret caret = tablature.getCaret();
+			control.requestCaretVerticalScroll();
+			if (caret.movePageDown(control.getVisibleHeight())) {
```
guiv42, line 26: I'm still not comfortable with this reference to a geometrical attribute of the displayed items (`visibleHeight`) at action level.
Better leave `Caret` evaluate the amplitude of the required movement: `caret.movePageDown()` (without argument).
(same remark for page up)

Note: I haven't really looked at the best way to retrieve this information from `Caret` implementation.

desktop/TuxGuitar/src/app/tuxguitar/app/view/component/tab/Caret.java
```diff
@@ -259,6 +260,131 @@ public void moveLeft() {
...
+		}
+
+		float currentY = this.selectedMeasure.getPosY();
+		Float targetY = null;
+		float expectedY = currentY + (direction * verticalDistance);
+		float bestYDistance = Float.MAX_VALUE;
+		boolean targetWithinPage = false;
+
+		/*
+		 * For line movement, choose the nearest distinct Y coordinate. For page
```
guiv42, line 306: See my preceding remark about renaming `verticalDistance` parameter: since there is a difference of logic between *line* and *page*, that would probably be better if the name of the parameter makes this explicit.
I don't know if I'm clear...

It's just that the logic [`verticalDistance==0` means *line*] is somehow implicit.

desktop/build-scripts/common-resources/common-macosx/Contents/MacOS/dist/tuxguitar-shortcuts.xml
guiv42, line 1: Could you please update the corresponding user documentation (default shortcuts)? It's there:
https://github.com/helge17/tuxguitar/blob/master/desktop/TuxGuitar/share/help/tools_shortcuts.html

I can also do it after merging if you prefer.

desktop/TuxGuitar/dist/tuxguitar-shortcuts.xml
```diff
@@ -17,6 +17,10 @@
 	<shortcut keys="-" action="action.note.duration.decrement-duration"/>
 	<shortcut keys="Up" action="action.caret.go-up"/>
 	<shortcut keys="Down" action="action.caret.go-down"/>
+	<shortcut keys="Alt Up" action="action.caret.go-line-up"/>
+	<shortcut keys="Alt Down" action="action.caret.go-line-down"/>
+	<shortcut keys="PgUp" action="action.caret.go-page-up"/>
```
guiv42, line 22: Replace "PgUp" by "Page Up"
(and similarly for "down")

I know that this is definitely not obvious...
There are currently some strings directly hard-coded in [SWT implementation](https://github.com/eclipse-platform/eclipse.platform.swt/blob/36ea085d73c789323455ffca86473193fa31b98d/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/MenuItem.java#L1891), at least for Linux, that led to some issue in the past (refer to #655 if you want more details)

desktop/build-scripts/common-resources/common-macosx/Contents/MacOS/dist/tuxguitar-shortcuts.xml
```diff
@@ -16,6 +16,10 @@
 	<shortcut keys="-" action="action.note.duration.decrement-duration"/>
 	<shortcut keys="Up" action="action.caret.go-up"/>
 	<shortcut keys="Down" action="action.caret.go-down"/>
+	<shortcut keys="Alt Up" action="action.caret.go-line-up"/>
+	<shortcut keys="Alt Down" action="action.caret.go-line-down"/>
+	<shortcut keys="PgUp" action="action.caret.go-page-up"/>
```
guiv42, line 21: See remark below: replace "PgUp" by "Page Up" (and similarly for "down")

I'm not 100% sure the SWT constraint is also valid for macOS, and I cannot test it. But I think it's still better if both xml files are aligned

guiv42 (commented):
Thanks for the update, and sorry for the review delay, the personal agenda is a bit complex these days.

When in linear mode, the keyboard shortcuts are ineffective, I guess it's intentional. Questions:
- page up/down: #1020 suggested to move 1 page left/right (logically: forwards/backwards), that seemed an interesting option to me

- line up/down:
  - in linear / single track mode I think it's OK (not a really useful mode)
  - in linear / multi-track mode, it could possibly be used to switch from one track to another

What do you think?

desktop/TuxGuitar/src/app/tuxguitar/app/view/component/tab/TGControl.java
woody6402: changed

desktop/TuxGuitar/src/app/tuxguitar/app/action/impl/caret/TGGoPageDownAction.java
woody6402: no better idea yet

desktop/build-scripts/common-resources/common-macosx/Contents/MacOS/dist/tuxguitar-shortcuts.xml
woody6402: done

desktop/TuxGuitar/dist/tuxguitar-shortcuts.xml
woody6402: Not sure if that's correct, but that is my quick view on that ....

.. we cannot replace PgUp/PgDn directly in the XML files at the moment:
- The shortcut parser currently uses spaces as key separators, so Page Up would be interpreted as two separate keys (Page + Up). In addition, the platform key mappings currently expose these keys as PgUp and PgDn via UIKey.
- Therefore, the XML keeps PgUp/PgDn for functional compatibility, while the user documentation displays the more
  readable Page Up/Page Down names.

desktop/build-scripts/common-resources/common-macosx/Contents/MacOS/dist/tuxguitar-shortcuts.xml
woody6402: see above

desktop/TuxGuitar/src/app/tuxguitar/app/view/component/tab/Caret.java
woody6402: tried to make it more clear
- by using definitive PageUp/Down and moveLineUp/Down
```
 	public boolean moveLineDown() {
-		return this.moveVertical(1, 0f);
+		return this.moveLine(1);
 	}
 
 	public boolean movePageUp(float visibleHeight) {
-		return this.moveVertical(-1, visibleHeight);
+		return this.movePage(-1, visibleHeight);
 	}
```
- and by changing the interface parameters
```
-	private boolean moveVertical(int direction, float verticalDistance) {
+	private boolean moveVertical(int direction, float movementDistance, boolean pageMovement) {
```

desktop/TuxGuitar/src/app/tuxguitar/app/view/component/tab/Caret.java
woody6402: changed

woody6402:
> When in linear mode, the keyboard shortcuts are ineffective, I guess it's intentional. Questions:
> 
> * page up/down: #1020 suggested to move 1 page left/right (logically: forwards/backwards), that seemed an interesting option to me

... could be done and would make additioal sense, but I would handle that as separete change (but see below)

> * line up/down:
>   
>   * in linear / single track mode I think it's OK (not a really useful mode)

maybe you can define:
- lineUp/Down for one shift left/right  (one line is then the visible tab line of screen width)
- and pageUp/Down multiple lines
- this proposal would be symetric to the behaviour in other modes

>   * in linear / multi-track mode, it could possibly be used to switch from one track to another
> 
> What do you think?

- not sure if track changing does make sense via line-up/down and/or page up/down. If you follow the idea of a cursor which is used for editiing or viewing then the one above reflects more this idea 

P.S.: also my answers will be delayed due to personal reasons ....

desktop/TuxGuitar/dist/tuxguitar-shortcuts.xml
guiv42: Yes you're right.
Then we have a (minor) conflict with TG shortcut parser and SWT menu item implementation for Linux. In other words, issue #655 is not fully solved: shortcuts "PgUp" and "PgDn" are not displayed in menus when they are associated to menu actions (confirmed by a test). No consequence here, since cursor movement actions are not accessible through menu items.

So let's keep your implementation unchanged, I'll handle the shortcuts/menu issue separately.

PR #1187, opened 2026-09-06
Export beat text in MusicXML (#289)

AzazelSensei:
Beat -> Insert Text was missing from the MusicXML export.

I write it as a direction/words element on the score staff, once per beat. The tab staff is left alone so the text is not duplicated after the backup.

See #289


guiv42 (approved):
OK, thanks for the contribution

Issue #984, opened 2026-01-14
Feature Request - Adding preview/remenancies of played notes inside the current bar


seleniumlux-afk:
Hi there, 

I thought a feature that already exists on GP could help people like me learning visually on the fretboard displayed. 
The idea is to have the fretboard displaying the full measure's pattern by highlighting all upcoming and past notes of the current measure in a faded (or other) color, providing visual anticipation.

<img width="932" height="905" alt="Image" src="https://github.com/user-attachments/assets/8839fbda-709d-4899-9617-3183d21caca7" />

guiv42:
First analysis: it may not be simple to draw a thin (empty) circle like suggested by the screenshot above.
The [UIPainter](https://github.com/helge17/tuxguitar/blob/master/common/TuxGuitar-lib/src/main/java/app/tuxguitar/ui/resource/UIPainter.java) interface does not offer this possibility currently, and it's implemented by several independent classes (for SWT/JFX/PDF/SVG...).

I can see 2 simpler options:
1. implement this new feature as a `default` method in the interface, only relying on other methods currently implemented by all instances (e.g. `cubicTo`). Simple, but not really clean.
2. use transparency parameter (call `setAlpha()`) to distinguish between current beat / other beats in the same measure: notes of current beat opaque / other notes slightly transparent

matevoux:
please check https://github.com/helge17/tuxguitar/discussions/1160 .
this is another way to do similar thing, note shape are moving at tempo.

For you're request may be possible tu use chord if on the score, if not when two chord on the same bar it will be messy

guiv42:
@seleniumlux-afk , @Snus2k, can you please elaborate a bit about your expectations?

More precise questions:

Q1: do you expect all measures notes to be shown *only while player is running*, or *permanently*?

Q2: I suggested to use transparency to distinguish between *current note* and *other notes in the measure*, both using the same color. Another suggestion was done to add a dedicated color for *other notes in the measure* (#1192). My personal opinion: too many different colors might create confusion (there are already 3 possible colors for notes on the fretboard). But my personal opinion is not so valuable, since I never use the feature myself.

seleniumlux-afk:
Hi @guiv42 ,

Thank you for your reply. Here are the ideas : 
[...]

PR #1192, opened 2026-09-12
Draw/Preview all notes on the fretboard in the current measure during playback

rajter:
Draw/Preview all notes on the fretboard in the current measure during playback or draw/preview all notes on the fretboard of the current selection.

https://github.com/user-attachments/assets/a32fa26f-db7b-41f3-b419-517185b97d68

Feature request and discussion:
https://github.com/helge17/tuxguitar/discussions/1106
https://github.com/helge17/tuxguitar/issues/984

### DIFF
--- a/common/resources/lang/messages.properties
+++ b/common/resources/lang/messages.properties
@@ -305,6 +305,7 @@ fretboard.background-color=Background Color
 fretboard.string-color=String Color
 fretboard.fretpoint-color=Fret Color
 fretboard.note-color=Note Color
+fretboard.note-measure-color=Note Color (Measure)
 fretboard.scale-note-color=Scale Note Color
 fretboard.tonic-color=Tonic Note Color
 fretboard.direction=Direction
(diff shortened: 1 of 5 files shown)


guiv42:
Before reviewing the code, I'd like to discuss a bit more about the *expected* behavior.
There are already 3 different colors for notes on the fretboard, I fear a little bit that adding another one will turn the fretboard into a rainbow. That may not be the best solution for readability. Please follow the discussion in #984

rajter:
I looked at the discussion #984 . We should definitely create a setting for the feature.
As for the coloring, i agree that adding another color could turn the fretboard into a rainbow like you said. But that is up to the user. I'm fine either way. 

Regarding the drawing, i made two examples
1. Only the outline
Fretboard_Outline

2. With transparency
Fretboard_Alpha

I set the alpha value to be 125. If we agree on this solution, should we make a setting for the alpha value or make it a fixed value that we agree upon?

